PR #2, opened 2026-08-11
Remove legacy Gateway provider seams

fazxes:
## Summary

- Remove the remaining legacy Gateway provider, schema, transport, and compatibility seams.
- Delete dead Gateway wrappers after their consumers move to the neutral boundary.
- Keep command metadata static and format provider labels at CLI and TUI presentation edges.
- Preserve Vercel authentication, routing, recovery, usage, and search behind the registered adapter with no fallback or second provider.

Stacked on #55.


fazxes:

PR #47, opened 2026-08-15
Introduce provider contracts and model descriptors

fazxes:
Stack: 1 of 10.
Base layer of the Gateway modularization stack.
Next: #48.

Summary
- Define neutral model request, stream event, cancellation, usage, and provider-tool contracts.
- Normalize model descriptors and keep Vercel-specific capability projection at the adapter edge.
- Keep native and WASM hosts on the same adapter contract while preserving the JS-host stream.

Verification
- Debug and ReleaseSafe tests
- Native, core-WASM, and terminal-WASM builds
- Terminal SDK flow
- Fresh repository binary smoke

PR #48, opened 2026-08-15
Persist Gateway connections and recovery routes

fazxes:
Stack: 2 of 10.
Depends on #47.
Next: #49.

Summary
- Add connection registry, route snapshots, and durable route authority.
- Keep selected credentials and resumed work pinned to the admitted connection.

Stack note
- Reuse the already-adopted selected credential after a failed logout so this slice remains independently functional.
- #52 replaces that compatibility bridge with exact adapter-owned resolution and removes it.

Verification
- state-machine and persistence tests
- zig build test
- connection, recovery, and failed-logout E2E flows


fazxes:

PR #49, opened 2026-08-15
Pin internal workloads to parent connections

fazxes:
Stack: 3 of 10.
Depends on #48.
Next: #50.

Summary
- Configure reviewer, vision, and subagent models per connection profile.
- Keep internal and resumed child work on the parent route without fallback.

Verification
- zig build test
- reviewer, vision, and subagent integration flows
- fresh binary recovery flow


fazxes:

PR #53, opened 2026-08-15
Move Vercel modules behind gateway ownership

fazxes:
Stack: 7 of 10.
Depends on #52.
Next: #54.

Summary
- Relocate Vercel OAuth, credential, JSON, and diagnostics modules under src/gateway.
- Keep the move mechanical with imports updated and behavior unchanged.

Verification
- direct-write ownership audit
- zig build test
- zig build
- fresh binary help and status smoke

PR #54, opened 2026-08-15
Define adapter-owned authentication contracts

fazxes:
Stack: 8 of 10.
Depends on #53.
Next: #55.

Summary
- Add bounded source inventory, entered-secret, presentation, and durable-write contracts.
- Pin profile identity and exact-origin credential adoption without adding another store owner.

Verification
- zig build test
- allocation and identity state tests
- fresh binary help and status smoke


fazxes:

PR #107, opened 2026-08-18
Omit catalog maxOutputTokens when it copies the context window

### LINKED ISSUE #106: Bug: Muse Spark 1.2 Contributor returns HTTP 400 invalid parameters
## Issue details

Selecting `meta/muse-spark-1.2-contributor` in interactive fx and sending a simple prompt (`hi`) fails with:

```
⚠ API request failed · HTTP 400 · AI_APICallError: The request contains invalid parameters. Check the request
  body for any errors or inconsistencies.
```

`fx ask` on the same model can succeed. Pi against the same Vercel AI Gateway also succeeds.

### Expected

Muse Spark 1.2 Contributor should complete the turn the same way other catalog models do.

### Steps

1. `fx login` and `/model meta/muse-spark-1.2-contributor`
2. Wait for the model catalog to load (interactive session)
3. Send `hi`

### Cause

After the catalog loads, fx copies `max_tokens` into `maxOutputTokens`. For Muse Spark the catalog advertises `max_tokens = context_window = 1048576`. Meta rejects that generation limit. Interactive sessions have the catalog ready; `fx ask` often sends the request before that metadata is applied.

### Sensitive information

Chinteyley:
Fixes #106.

Interactive fx fails on `meta/muse-spark-1.2-contributor` with HTTP 400 `AI_APICallError: The request contains invalid parameters`. After the Gateway catalog loads, fx sends `maxOutputTokens: 1048576` because the catalog copies the 1M context window into `max_tokens`. Meta rejects that generation limit. Pi works on the same gateway because it does not send that field.

Treat a catalog output limit that is greater than or equal to the context window as unset, so Muse and similar models omit `maxOutputTokens` and keep using real caps such as 32k on a 256k window.

Required label: `type: bug`.

### DIFF
--- a/src/builtins/gateway.zig
+++ b/src/builtins/gateway.zig
@@ -304,6 +304,29 @@ test "agent request builder keeps default reasoning silent and emits output limi
     try std.testing.expect(std.mem.find(u8, body, "\"providerOptions\"") == null);
 }
 
+test "agent request builder omits output limit when catalog copies the context window" {
+    const messages = [_]shared_types.ChatMessage{.{ .role = .user, .content = "question" }};
+    const capabilities = model_capabilities.resolveCapabilities("meta/muse-spark-1.2-contributor", .{
+        .context_window = 1_048_576,
+        .max_output_tokens = 1_048_576,
+    });
+    const body = try agent_stream_provider.build(std.testing.allocator, .{
+        .model = "meta/muse-spark-1.2-contributor",
+        .serialized_tools = "[]",
+        .messages = &messages,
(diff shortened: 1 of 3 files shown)


fazxes:
Thanks @Chinteyley for the clear report, diagnosis, and patch. While validating the fix, I ended up testing and merging #119 instead. #119 fixes the same bug at the request boundary, preserving the catalog metadata while omitting unusable full-window limits from agent requests. Since that fix is now on main, I’m closing this PR as superseded. Sorry for the mix-up, and thank you for catching this.

PR #120, opened 2026-08-19
Add Nix flake

y0usaf:
## Why

Let Nix users install fx through Nix.

## What changed

- Adds a package-only flake for Linux and macOS on x86_64 and ARM64.
- Disables automatic updates because Nix owns the installed files.
- Documents `nix profile add github:vercel-labs/fx`.

## Pin choice

The lock uses the final nixpkgs branch supporting Intel macOS, matching fx's platform support without another package source.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,7 +18,12 @@
               in
               pkgs.lib.removeSuffix "\";" (pkgs.lib.removePrefix "pub const version = \"" line);
             src = ./.;
+            # Build/install phases come from the nixpkgs zig setup hook, which
+            # runs `zig build` / `zig build install` when build.zig is present.
             nativeBuildInputs = [ pkgs.makeBinaryWrapper pkgs.zig ];
+            doInstallCheck = true;
+            nativeInstallCheckInputs = [ pkgs.versionCheckHook ];
+            versionCheckProgramArg = "--version";
             postInstall = ''
               install -Dm444 LICENSE "$out/share/licenses/fx/LICENSE"
               install -Dm444 THIRD_PARTY_NOTICES.md "$out/share/licenses/fx/THIRD_PARTY_NOTICES.md"
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,13 @@
         in
         {
           default = pkgs.stdenv.mkDerivation {
-            name = "fx";
+            pname = "fx";
+            version =
+              let
+                line = pkgs.lib.findFirst (pkgs.lib.hasPrefix "pub const version = ") null
+                  (pkgs.lib.splitString "\n" (builtins.readFile ./src/main.zig));
+              in
+              pkgs.lib.removeSuffix "\";" (pkgs.lib.removePrefix "pub const version = \"" line);
             src = ./.;
             nativeBuildInputs = [ pkgs.makeBinaryWrapper pkgs.zig ];
             postInstall = ''
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,7 @@
               install -Dm444 LICENSE "$out/share/licenses/fx/LICENSE"
               install -Dm444 THIRD_PARTY_NOTICES.md "$out/share/licenses/fx/THIRD_PARTY_NOTICES.md"
             '';
-            postFixup = "wrapProgram $out/bin/fx --set FX_AUTO_UPGRADE 0";
+            postFixup = "wrapProgram $out/bin/fx --set-default FX_AUTO_UPGRADE 0";
             meta = {
               license = pkgs.lib.licenses.asl20;
               mainProgram = "fx";
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,23 @@
+{
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-26.05-darwin";
+
+  outputs = { nixpkgs, ... }:
+    let
+      systems = [ "aarch64-darwin" "aarch64-linux" "x86_64-darwin" "x86_64-linux" ];
+      eachSystem = f: nixpkgs.lib.genAttrs systems (system: f nixpkgs.legacyPackages.${system});
+    in
+    {
+      packages = eachSystem (pkgs: {
+        default = pkgs.stdenv.mkDerivation {
+          name = "fx";
+          src = ./.;
+          nativeBuildInputs = [ pkgs.makeBinaryWrapper pkgs.zig_0_16 ];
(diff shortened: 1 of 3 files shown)


y0usaf:
I trimmed this back to a package-only flake. It should stay useful to Nix users without becoming another thing for the Vercel team to maintain.

flake.nix
```diff
@@ -0,0 +1,28 @@
...
+  outputs = { nixpkgs, ... }: {
+    packages = nixpkgs.lib.genAttrs
+      [ "aarch64-darwin" "aarch64-linux" "x86_64-darwin" "x86_64-linux" ]
+      (system:
+        let
+          pkgs = nixpkgs.legacyPackages.${system};
+        in
+        {
+          default = pkgs.stdenv.mkDerivation {
+            name = "fx";
```
keepinfov: `name` gives a store path and `nix profile list` entry with no version. Deriving it from the same `pub const version` line that `release.yml` greps keeps it correct without anyone having to remember this file at release time.

```suggestion
            pname = "fx";
            version =
              let
                line = pkgs.lib.findFirst (pkgs.lib.hasPrefix "pub const version = ") null
                  (pkgs.lib.splitString "\n" (builtins.readFile ./src/main.zig));
              in
              pkgs.lib.removeSuffix "\";" (pkgs.lib.removePrefix "pub const version = \"" line);
```

flake.nix
```diff
@@ -0,0 +1,28 @@
...
+            src = ./.;
```
keepinfov, line 20: `./.` copies the whole tree, so `tests/`, `sdk/` and `.github/` land in the store and any docs-only commit invalidates the build. Mirroring `.paths` from `build.zig.zon` takes the source closure from 30.5 MiB to 25.5 MiB here.

```suggestion
            src = pkgs.lib.fileset.toSource {
              root = ./.;
              fileset = pkgs.lib.fileset.unions [
                ./build.zig
                ./build.zig.zon
                ./src
                ./LICENSE
                ./THIRD_PARTY_NOTICES.md
              ];
            };
```

flake.nix
```diff
@@ -0,0 +1,28 @@
...
+            nativeBuildInputs = [ pkgs.makeBinaryWrapper pkgs.zig ];
```
keepinfov, line 23: `pkgs.zig` and `pkgs.zig_0_16` are the same derivation on your pin today, but `zig` follows nixpkgs, so the next lock bump moves it off the `minimum_zig_version` in `build.zig.zon`. `versionCheckHook` also makes the build fail if the installed binary stops reporting the version parsed above.

```suggestion
            nativeBuildInputs = [ pkgs.makeBinaryWrapper pkgs.zig_0_16 ];
            doInstallCheck = true;
            nativeInstallCheckInputs = [ pkgs.versionCheckHook ];
            versionCheckProgramArg = "--version";
```

flake.nix
```diff
@@ -0,0 +1,28 @@
...
+            postInstall = ''
+              install -Dm444 LICENSE "$out/share/licenses/fx/LICENSE"
+              install -Dm444 THIRD_PARTY_NOTICES.md "$out/share/licenses/fx/THIRD_PARTY_NOTICES.md"
+            '';
+            postFixup = "wrapProgram $out/bin/fx --set FX_AUTO_UPGRADE 0";
```
keepinfov: `--set` makes this impossible to override. `--set-default` still disables automatic updates by default but leaves the escape hatch.

```suggestion
            postFixup = "wrapProgram $out/bin/fx --set-default FX_AUTO_UPGRADE 0";
```

flake.nix
```diff
@@ -0,0 +1,28 @@
...
+            meta = {
+              license = pkgs.lib.licenses.asl20;
+              mainProgram = "fx";
+            };
```
keepinfov, line 38: `description` and `homepage` show up in `nix search`, and `platforms` keeps the package from evaluating on systems fx does not ship.

```suggestion
            meta = {
              description = "Tiny, open, embeddable, native coding agent";
              homepage = "https://fx.sh";
              license = pkgs.lib.licenses.asl20;
              mainProgram = "fx";
              platforms = [ "aarch64-darwin" "aarch64-linux" "x86_64-darwin" "x86_64-linux" ];
            };
```

keepinfov (commented):
Built this on x86_64-linux and it works. Five optional suggestions inline. I applied all of them together and rebuilt on your pin, so they compose.

flake.nix
y0usaf: I _think_ this isn't necessary because nixpkgs is pinned by flake.lock, but I think this is a suitable suggestion, so feel free to correct me if I'm wrong regarding nixpkgs already being pinned, in which case I'm happy to accept.

flake.nix
y0usaf: IMO this adds maintenance burden, so I won’t accept it right away. Feel free to push back, but the goal of this PR is to make maintenance as frictionless as possible for the Vercel team.

y0usaf: IMO this adds maintenance burden, so I won't accept instantly. Feel free to push back, but the goal of the PR is to make maintenance as frictionless as possible for the Vercel team.

y0usaf:
Many thanks @keepinfov, I've left 2 of your suggestions open, let's hope 𝒇x gets Nix support :)

flake.nix
```diff
@@ -0,0 +1,37 @@
...
+          default = pkgs.stdenv.mkDerivation {
+            pname = "fx";
+            version =
+              let
+                line = pkgs.lib.findFirst (pkgs.lib.hasPrefix "pub const version = ") null
+                  (pkgs.lib.splitString "\n" (builtins.readFile ./src/main.zig));
+              in
+              pkgs.lib.removeSuffix "\";" (pkgs.lib.removePrefix "pub const version = \"" line);
+            src = ./.;
+            nativeBuildInputs = [ pkgs.makeBinaryWrapper pkgs.zig ];
```
DarlingtonDeveloper, line 23: **[P1] Add build and install phases**

This derivation never runs `zig build` and never installs a binary. With no Makefile, the generic phases do nothing; `postInstall` only installs license files, then `postFixup` attempts to wrap nonexistent `$out/bin/fx`. Add explicit build/install phases and a check that executes `$out/bin/fx --version`.

DarlingtonDeveloper (changes requested):

y0usaf:
@DarlingtonDeveloper Ready for re-review! Sorry about the random closing, call it agent noise :')

PR #156, opened 2026-08-19
Add Grok OAuth sign-in

christopherkarani:
## Summary

- Adds `fx login grok` / `fx logout grok` (and `/login grok` / `/logout grok`) so SuperGrok or X Premium+ can authenticate fx without an AI Gateway key.
- Stores that session in `~/.fx/grok-auth.json` beside Vercel `auth.json`. Optional `fx login grok --import` copies tokens from `~/.grok/auth.json` without writing it.
- When Grok OAuth is the selected credential, chat goes to `https://api.x.ai/v1/chat/completions` as OpenAI Chat Completions SSE, identified as `fx/<version>`. Grok tokens are never sent to AI Gateway.

## Test plan

- [x] `zig build`
- [x] Focused Zig tests for grok session, login parse, xAI JSON, catalog isolation
- [x] `cd tests/e2e && bun test grok-oauth.test.ts`
- [x] `./zig-out/bin/fx help` shows `login [grok] [--import]`
- [x] `./zig-out/bin/fx logout grok` reports no session
- [ ] Live `./zig-out/bin/fx login grok` against SuperGrok (operator)
- [ ] Full CI on this commit


src/gateway/xai_stream_provider.zig
```diff
@@ -0,0 +1,181 @@
...
+        .location = .{ .url = request.chat_url },
+        .method = .POST,
+        .payload = request.payload,
+        .headers = .{
+            .user_agent = .{ .override = gateway_client.user_agent },
+            .accept_encoding = .omit,
+            .content_type = .{ .override = "application/json" },
+            .authorization = .{ .override = auth_header },
+        },
+        .response_writer = &body.writer,
```
DarlingtonDeveloper, line 64: **[P1] Stream the response instead of buffering it**

`Client.fetch` drains the entire response into `body` before `consumeSse` runs. Content and tool callbacks therefore wait until xAI closes the response, cancellation cannot interrupt a stalled stream, and the allocating writer has no response-size bound. Feed a streaming reader to an incremental SSE decoder with a cancellation watcher.

src/gateway/xai_json.zig
```diff
@@ -0,0 +1,351 @@
...
+    if (objectString(choice.object, "finish_reason")) |reason| {
+        delta.finish_reason = try alloc.dupe(u8, reason);
+    }
+    const delta_val = choice.object.get("delta") orelse return delta;
+    if (delta_val != .object) return delta;
+    if (objectString(delta_val.object, "content")) |content| {
+        if (content.len > 0) delta.content = try alloc.dupe(u8, content);
+    }
+    if (delta_val.object.get("tool_calls")) |calls| {
+        if (calls == .array and calls.array.items.len > 0 and calls.array.items[0] == .object) {
```
DarlingtonDeveloper, line 220: **[P1] Preserve every streamed tool-call index**

The request enables `parallel_tool_calls`, but this parser only reads `calls[0]`, while `consumeSse` uses one global id/name/arguments accumulator and ignores `tool_call_index`. Multiple tool calls are dropped or merged. Accumulate every array entry by index and add a two-call regression.

PR #165, opened 2026-08-19
Keep terminal helpers alive after a nested rebuild

### LINKED ISSUE #166: Bug: terminal exec fails with FileNotFound after fx rebuilds itself
## What happens

I was using interactive `fx` in its own checkout. Tools worked at first. Mid-session the agent ran `zig build`, which replaced `./zig-out/bin/fx` on disk with a large Debug binary.

After that, later `terminal` calls failed with `FileNotFound`. The first command had succeeded. Restarting fx made tools work again until the next in-session rebuild.

Typical tool result:

```json
{"error":{"type":"tool_execution_failed","tool_name":"terminal","message":"Tool execution failed","details":{"error":"FileNotFound"}}}
```

The TUI showed rows like `Failed cd . && zig fmt ...` even though the same command had just worked.

## Expected

A nested `zig build` (or any command that replaces `zig-out/bin/fx`) should not break later `terminal exec` in the same process.

## How to reproduce

1. From this repo, start the interactive binary you just built: `./zig-out/bin/fx`
2. Use yolo or otherwise allow the `terminal` tool.
3. Ask fx to run `zig build` in the workspace (Debug is enough; that overwrites `zig-out/bin/fx`).
4. After that command exits 0, ask fx to run a trivial follow-up such as `ls` or `zig version`.

Result: the follow-up `terminal exec` fails with `FileNotFound` until you quit and start fx again.

I hit this on WSL2 Linux, fx 0.0.3, running `./zig-out/bin/fx` from the repo. Fix: #165

acking-you:
Fixes #166

## Summary

Interactive `fx` re-execs itself for `terminal` helpers. After a nested `zig build` replaces `zig-out/bin/fx` on disk, `executablePathAlloc` realpaths `/proc/self/exe` to that old path and later helper spawns return `FileNotFound`. Same-process re-exec now uses `/proc/self/exe`; later peer launches (tmux bootstrap) use `/proc/<pid>/exe`.

```text
before:  /proc/self/exe  → realpath → zig-out/bin/fx  → rebuild → ENOENT
after:   /proc/self/exe  or  /proc/<pid>/exe  → live inode, still runs
```

## Testing

- `zig fmt --check src/`
- Reproduced on WSL with the steps in #166; after this change the follow-up `terminal exec` keeps working
- Required label: `type: bug`


DarlingtonDeveloper (changes requested):
The `/proc/self/exe` approach looks reasonable, and I verified that the branch merges cleanly and cross-compiles for both Linux architectures. However, this needs Linux regression coverage before merging.

The current test does not exercise the production Linux branch during normal `zig build test`: `FX_TEST_PRODUCT_EXE` is set by the build, so the test returns after only checking that the override is non-empty. `pathForPeerReexec` also has no coverage, and this PR currently has only the two Socket checks rather than the project CI suite.

Please add a Linux regression test that replaces the on-disk fx binary and proves both same-process re-exec (`/proc/self/exe`) and peer re-exec (`/proc/<pid>/exe`) continue working, then run the normal project CI on the exact commit.

acking-you:
> The current test does not exercise the production Linux branch during normal `zig build test`: `FX_TEST_PRODUCT_EXE` is set by the build, so the test returns after only checking that the override is non-empty. `pathForPeerReexec` also has no coverage, and this PR currently has only the two Socket checks rather than the project CI suite.
> 
> Please add a Linux regression test that replaces the on-disk fx binary and proves both same-process re-exec (`/proc/self/exe`) and peer re-exec (`/proc/<pid>/exe`) continue working, then run the normal project CI on the exact commit.

Added the Linux regression coverage you asked for — the old test never
reached the production branch because build.zig always sets
FX_TEST_PRODUCT_EXE. Could you approve the workflows so CI can run on this
commit? @DarlingtonDeveloper

acking-you:
Hi @DarlingtonDeveloper, thank you again for the careful review.

I have rebased the branch onto the latest upstream main at 4a8f765, and GitHub now reports the PR as mergeable again. I also addressed the requested Linux coverage: the production re-exec branch is exercised directly despite FX_TEST_PRODUCT_EXE, the peer re-exec path is covered, and the focused E2E starts a live terminal host, replaces its on-disk fx binary, then verifies that both tmux and native helpers continue to run. The ReleaseSafe build and that focused Linux E2E both pass locally, so the reported bug should now be fixed on the current upstream base.

For context, I maintain a fork of fx where I experiment with some additional ideas and features, while regularly merging upstream, often daily. I think fx is an excellent project, and I would be very happy to participate and contribute more actively upstream as well.

I am tagging you again mainly because upstream is moving very quickly and the branch may accumulate more conflicts if it waits much longer, not because I want to put pressure on you. When you have time, could you please take another look and approve the workflows or re-review the updated PR? Thank you.

fazxes (approved):

PR #168, opened 2026-08-19
Add OpenAI-compatible client transport

boozedog:
fx can talk to any OpenAI-compatible server as a client. Vercel AI Gateway stays the default. With `OPENAI_API_KEY` (and optional `FX_OPENAI_BASE_URL`), Vercel login is not required.

## Usage

Chat Completions is the default OpenAI-compatible wire (`/v1/chat/completions`). That path works with Ollama, Groq, LiteLLM, vLLM, and most compatible hosts:

```bash
export OPENAI_API_KEY=your-key
export FX_MODEL=gpt-4o
# optional: export FX_OPENAI_BASE_URL=http://127.0.0.1:11434/v1
fx
```

For official OpenAI models that need the Responses API (tools plus reasoning on gpt-5.x), set the Responses wire explicitly:

```bash
export OPENAI_API_KEY=your-key
export FX_OPENAI_API_STYLE=responses
export FX_MODEL=gpt-5
fx
```

`~/.fx/settings.json` can store `openai_api_key`, `openai_base_url`, and `openai_api_style`. Project `.fx.json` cannot.

## Behavior

- Agent history still maps through the existing stream-provider contract. The Vercel AI SDK Gateway path is unchanged when OpenAI mode is off.
- Streaming tool calls work on both OpenAI wires.
- `GET /v1/models` is used when the server implements it; a missing catalog is not fatal.
- Gateway-only surfaces (credits, team picker, Gateway web search, fast mode, generation-usage follow-up) degrade instead of failing opaquely.
- Transport routing uses an explicit wire kind (`gateway`, `openai_chat`, `openai_responses`) rather than URL suffix sniffing, so a later native Anthropic adapter can add a variant instead of another boolean.
- No new package dependencies.

## Tests

- Zig unit coverage for request JSON, tool envelopes, SSE, and route construction
- `tests/e2e/openai-compatible-fake.test.ts` and `tests/e2e/openai-responses-fake.test.ts` (classified in `scripts/pgso/corpus.json`)

## Out of scope

Vision, structured output, OpenAI hosted tools, `previous_response_id` conversation state, and any third-party Zig client library.


src/gateway/openai_client.zig
```diff
@@ -0,0 +1,609 @@
...
+
+        req.transfer_encoding = .{ .content_length = request.payload.len };
+        var send_buf: [8192]u8 = undefined;
+        var body_writer = req.sendBodyUnflushed(&send_buf) catch return error.WriteFailed;
+        body_writer.writer.writeAll(request.payload) catch return error.WriteFailed;
+        body_writer.end() catch return error.WriteFailed;
+        req.connection.?.flush() catch return error.WriteFailed;
+        if (request.delivery) |delivery| delivery.markPossiblySent();
+
+        var response = req.receiveHead(&.{}) catch |err| {
```
DarlingtonDeveloper, line 230: **[P1] Make Chat Completions reads cancellable**

After the bounded open, `receiveHead` and the SSE reader perform blocking reads with only before/after flag checks. If a server stops sending, Escape cannot close the socket and the turn can hang indefinitely. Install the equivalent of the Responses path’s HTTP cancellation watcher around send, head, and body reads.

src/gateway/openai_client.zig
```diff
@@ -0,0 +1,609 @@
...
+                    content_capture_limit,
+                );
+            },
+            .done => break,
+            .ignored => continue,
+            .read_failed => {
+                if (cancel_flag.load(.seq_cst)) return error.Cancelled;
+                return error.ReadFailed;
+            },
+            .eof => break,
```
DarlingtonDeveloper, line 340: **[P1] Do not turn an incomplete EOF into success**

`.eof` is treated like `[DONE]`, then any partial content or tools are assigned `.stop`/`.tool_calls`. A connection closing after one delta is therefore committed as a successful turn without terminal evidence. Return an incomplete/read failure on bare EOF without a finish reason and add a content-delta-plus-EOF regression.

PR #172, opened 2026-08-19
Skip managed-child probes in ordinary session lists

IfkumRfnl:
## Summary
- skip managed-child relationship-index probes for ordinary session lists
- preserve probing for resumable and index-writing paths
- cover shallow and probing scan behavior with a focused regression test

## Local verification
- `zig fmt --check src/`
- focused managed-child scan test
- focused session-list test
- focused resumable-index backfill test
- `zig build`
- built-binary smoke: `sessions --json` over eight index-less sessions
- built-binary smoke: `background --json` over the same fixture


fazxes (approved):
Reviewed the session scan split and verified that ordinary listing paths skip only the managed-child relationship probe, while resumable and index-writing paths retain it.

Issue #158, opened 2026-08-19
Feature request: Option to rename terminal tabs with the session name


Jamir-boop:
## Feature request: Renameable terminal tabs mapped to session names

It would be very useful to be able to rename terminal tabs with the session name, so that when running multiple sessions in parallel it is easy to identify which tab corresponds to which session.

### Why

When working with several fx sessions at once, every tab looks the same (same default title) and it is hard to tell them apart at a glance. Letting each tab show the session name (or a user-defined label) would make navigation and context switching much faster.

### Proposed behavior

- A way to set a custom name/label per terminal tab that reflects the active session name.
- Ideally auto-updated when the session is (re)named, with an optional manual override for power users.
- Could be exposed as a tab context-menu action (right-click > Rename) and/or a CLI/flag option.

### Context

![Untitled](https://raw.githubusercontent.com/Jamir-boop/markdown-images/master/Untitled.png)

### Environment

- fx (recent build)
- WSL2 (Ubuntu 24.04) on Windows

Happy to help refine the UX or test an implementation if this is something the team would consider.

PR #174, opened 2026-08-19
Name terminal tabs after the active session

arthjean:
Refs #158

## Problem

Every fx tab renders the same title, `fx · <model>`, so running several sessions in parallel gives no way to tell them apart at a glance.

## Behavior

The terminal title carries the session name once one exists, and falls back to the model before the first turn names the session:

| Moment | Title |
| --- | --- |
| Startup, no session name yet | `fx · openai/gpt-5` |
| After the first turn derives a name | `fx · generate the release notes` |
| After `/rename deploy pipeline fix` | `fx · deploy pipeline fix` |
| After a model switch on a named session | unchanged |
| After `/clear` or a new session | back to `fx · <model>` |
| `fx resume <id>` / `fx session resume <id>` | `fx · deploy pipeline fix` |
| Exit | restored to the terminal's own title |

`/rename` doubles as the manual override the issue asks for: renaming the session renames the tab. The right-click rename the issue mentions belongs to the terminal emulator and is outside what fx can drive.

## Implementation

`Runtime(App).syncTerminalTitle` is the single place that resolves "session name, else model", so every mutation point routes through it: the derived title, `/rename`, the resume path, session reset, and model switches.

Two correctness points came out of the work:

* The provider wrote to `std.Io.File.stdout()` directly. It now writes to the host's output file (`terminalTitleFor(&shell.stdout_file)`), so a host that redirects its output keeps the escape sequence out of the real stdout. This also unblocks `zig build test`, where stray stdout bytes deadlock the `--listen=-` protocol channel.
* Titles are sanitized of C0 and DEL bytes at the cache boundary. They are derived from prompt text and from the display sidecar, so neither source is trusted. Verified with a forged `display.json` title of `safe\x07\x1b]2;PWNED\x7ftail`, which renders as `fx · safe]2;PWNEDtail`.

The host capability was renamed from `setModel`/`set_model_fn` to `set`/`set_fn`, since the label is no longer model-specific.

## Tests

* `src/ui/render.zig`: the provider writes the label to the caller's output file.
* `src/core/app/app_session_runtime.zig`: the title follows the session name and falls back to the model; control bytes are dropped before reaching the terminal.
* `src/core/hosts/host.zig`: the capability forwards borrowed label bytes and clear.
* `tests/e2e/tui-slash-menu.test.ts`: `terminal tab title follows the session name across rename and resume` drives the real binary in tmux across startup, derived name, `/rename`, `/quit`, and resume.
* `TmuxSession.paneTitle()` replaces a local helper that ignored isolated tmux sockets.

## Docs

`README.md` gains one paragraph in the session block, since the behavior is user-facing.

## CI

Full CI triggers on `push` to a non-`main` branch, so it runs in the fork: https://github.com/arthjean/fx/actions?query=branch%3Aarthjean%2Fname-terminal-tabs-after-session

The `pull_request` workflows on this repository report `action_required` and need a maintainer to approve workflow runs for this fork PR.

## Label

Requires `type: improvement`. The tab title already existed and already showed the model; this changes what it displays rather than adding a new surface. I cannot apply labels on this repository, so the PR stays in draft until a maintainer applies one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ fx session resume last
 fx session resume --id <id>
 ```
 
+Each session names its terminal tab. The title carries the session name once one exists, so parallel fx tabs stay distinguishable, and falls back to the active model before the first turn names the session. Renaming a session with `/rename` renames its tab.
+
 Run `/feedback` to open the feedback form at `fx.sh/feedback`. It does not create a diagnostic or change the clipboard.
 
 Run `/trace` to create a private Markdown diagnostic with logs, session context, runtime state, permissions, and recent activity. On macOS, fx copies the `.md` file to the clipboard; on other platforms, it saves the file and prints its path. Review and redact the trace before sharing it.
--- a/src/core/app/app_bootstrap_runtime.zig
+++ b/src/core/app/app_bootstrap_runtime.zig
@@ -437,7 +437,7 @@ pub fn Runtime(comptime App: type) type {
             if (app.requested_resume == null) {
                 try deps.begin_fresh_persisted_session(app);
                 deps.enable_session_stores(app);
-                deps.terminal_title.setModel(app.selected_model.items);
+                deps.terminal_title.set(app.selected_model.items);
             }
 
             switch (staged_resume_view) {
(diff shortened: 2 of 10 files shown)


DarlingtonDeveloper (approved):

DarlingtonDeveloper:
Superseded by #177, which carries this implementation forward in a verified signed commit and adds the FXC-232 session/workspace/model context changes. Thank you @arthjean — co-author credit is preserved in the replacement commit.

arthjean:
Thanks for the credit, @DarlingtonDeveloper. Noted on signed commits. I'll sign from now on so the next one can land directly.

PR #180, opened 2026-08-20
Allow external symlink authorities for skill discovery

### LINKED ISSUE #132: Skills installed via Nix symlinks (outside home) cannot be loaded
## Summary

fx follows symlinks inside skill directories, but rejects any link whose resolved target falls outside the home directory (or workspace root). When a CLI tool installed via Nix provides skills, the actual files live under `/nix/store/...` and only a symlink is placed at `~/.claude/skills/<name>` (or similar). fx cannot discover or load these skills.

## Steps to reproduce

1. Install a CLI tool via Nix that ships skills (e.g. [hunk](https://github.com/modem-dev/hunk))
2. The tool creates a symlink at `~/.claude/skills/<name>` pointing to `/nix/store/...`
3. Start fx
4. The skill is not recognized

## Expected behavior

Skills reachable via symlinks that point outside home (e.g. to the Nix store) should still be discovered and loaded.

## Actual behavior

The skill is skipped during discovery and does not appear in the catalog.

## Root cause

The `openContainedDir` function introduced in [#83](https://github.com/vercel-labs/fx/pull/83) (`src/core/skills/skill_runtime.zig`) resolves symlinks via `realpathAlloc` and then checks `pathInside(read_authority, canonical_path)`. When the resolved target is outside `read_authority` (home or workspace_root), it returns `PathOutsideReadAuthority` and the candidate is skipped.

Additionally, `root_policy` is hardcoded in `src/builtins/skills.zig`. There is no setting or environment variable to relax this boundary.

## Affected use cases

- Users managing skills via Nix / home-manager
- Skills installed with `npx skills add` whose symlinks resolve to the Nix store
- CLI tools like hunk that assume symlink-based skill installation and are installed via Nix

## Environment

- fx v0.0.3
- macOS (Darwin 25.3.0, Apple Silicon)
- Nix package manager

## Proposal

One of the following approaches would unblock this use case:

1. Allow extending `read_authority` boundaries via config file or environment variable
2. Add an allowlist for known-safe external paths (e.g. `/nix/store/`)
3. Provide an opt-in flag to follow symlinks outside the authority boundary

## Willing to implement

I'd be happy to take this on if the team is open to any of the above approaches. Since this touches the security boundary, I'd like to align on the direction first, then open a draft PR.

yamad07:
Fixes #132

## Summary

Skills installed via Nix or similar package managers place symlinks inside `~/.claude/skills` (or other skill roots) that resolve to paths outside the home directory, such as `/nix/store/...`. The `openContainedDir` function rejected these symlinks because the resolved target fell outside `read_authority` (home or `workspace_root`), causing the skill to be silently skipped during discovery.

## Changes

- Add `FX_SKILL_SYMLINK_AUTHORITIES` environment variable: a colon-separated list of absolute paths that extend the set of trusted directories for symlink resolution
- When a skill symlink resolves outside the owning authority, each configured external authority is checked before rejecting the candidate
- Relative entries and entries containing `..` components are silently skipped
- Update README to document the new environment variable

## How it works

`openContainedDir` in `src/core/skills/skill_runtime.zig` now:
1. Resolves the symlink via `realpathAlloc` as before
2. Checks if the canonical path is inside `read_authority` (home or workspace) — if so, opens it as before
3. If not, parses `FX_SKILL_SYMLINK_AUTHORITIES` and checks each configured authority
4. Only if no authority matches does it return `PathOutsideReadAuthority`

This is the same opt-in allowlist approach proposed in the issue (option 2), keeping the security boundary explicit and user-controlled.

## Usage

```bash
# Allow skills whose symlinks resolve into the Nix store
export FX_SKILL_SYMLINK_AUTHORITIES=/nix/store

# Multiple authorities
export FX_SKILL_SYMLINK_AUTHORITIES=/nix/store:/opt/skills
```

## Test plan

- [x] `zig build` succeeds
- [x] `zig fmt --check src/` passes
- [x] Unit test: `externalSymlinkAuthorities parses colon-separated absolute paths`
- [x] Unit test: `externalSymlinkAuthorities returns empty when unset`
- [x] Unit test: `loadVisibleSkills discovers a linked candidate resolved via external symlink authority`
- [x] Unit test: `loadVisibleSkills still rejects external symlinks without an authority`
- [ ] Full CI on all four native runners

Note: 6 pre-existing test failures in the local environment are caused by a zsh `zle` configuration issue (`(eval):1: can't change option: zle` polluting stderr) and are unrelated to this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ The WebAssembly SDK is experimental. See the [WebAssembly SDK](sdk/README.md) an
 
 ## Extend fx
 
-Add reusable instructions with [skills](https://fx.sh/docs/capabilities/skills), connect external tools through [MCP](https://fx.sh/docs/capabilities/mcp), or delegate independent work to [subagents](https://fx.sh/docs/capabilities/subagents). Project instruction files may link within their scope, and read-only workspace or compatibility skill directories may link within their owning workspace or home; managed skills, `SKILL.md` files, resources, and escaping links remain no-follow. `fx status` and `fx doctor` report an invalid trusted MCP profile without starting its servers.
+Add reusable instructions with [skills](https://fx.sh/docs/capabilities/skills), connect external tools through [MCP](https://fx.sh/docs/capabilities/mcp), or delegate independent work to [subagents](https://fx.sh/docs/capabilities/subagents). Project instruction files may link within their scope, and read-only workspace or compatibility skill directories may link within their owning workspace or home; managed skills, `SKILL.md` files, resources, and escaping links remain no-follow. Skills installed via symlinks that resolve outside home or workspace (e.g. Nix store paths) are loaded when their resolved target is inside a directory listed in the `FX_SKILL_SYMLINK_AUTHORITIES` environment variable (colon-separated absolute paths). `fx status` and `fx doctor` report an invalid trusted MCP profile without starting its servers.
 
 ## Documentation
 
--- a/src/core/skills/skill_runtime.zig
+++ b/src/core/skills/skill_runtime.zig
@@ -379,8 +379,62 @@ fn openContainedDir(
 ) !std.Io.Dir {
     const canonical_path = try io_mod.realpathAlloc(alloc, logical_path);
     defer alloc.free(canonical_path);
-    if (!pathing.pathInside(read_authority, canonical_path)) return error.PathOutsideReadAuthority;
-    return io_mod.openDirAbsoluteNoFollow(canonical_path, options);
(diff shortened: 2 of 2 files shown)


DarlingtonDeveloper:
Superseded by #199, which carries the patch-identical implementation as a verified signed commit on current `main`. All required checks passed, including Full CI across Linux and macOS, and #199 has merged. #132 is now closed as completed. Closing this PR.

yamad07:
Thank you!!

PR #212, opened 2026-08-20
Stop undo from destroying files it cannot restore

### LINKED ISSUE #211: Undo can destroy a file and report it as restored
### Summary

Undo can destroy a workspace file and report the undo as successful. There are two independent paths into it, both in `src/core/workspace/change_tracker.zig`.

### 1. A failed restore write is swallowed after the file is already truncated

`ChangeTracker.undoLast` restores a preimage with:

```zig
var file = std.Io.Dir.createFileAbsolute(io_mod.getIo(), op.path, .{ .truncate = true }) catch { ... };
defer file.close(io_mod.getIo());
file.writeStreamingAll(io_mod.getIo(), content) catch {};
alloc.free(content);
return .{ .restored = op.path };
```

Opening with `.truncate = true` destroys the file's bytes before anything is written. If `writeStreamingAll` then fails (ENOSPC, EFBIG, a quota, an IO error), the error is discarded, `content` (the only in-memory copy of the preimage) is freed, and the function still returns `.restored`, so `/undo` prints `Restored <path>`. The file is left truncated or half written with no way to recover it.

The same shape appears three times: the delete-restore branch, the rename destination-preimage branch, and the write/edit branch.

Reproduced by lowering `RLIMIT_FSIZE` to 4 KiB with a 64 KiB preimage: `undoLast` returned `.restored` and left the target at 4096 bytes, neither the 46-byte original nor the 64 KiB preimage.

### 2. "Could not read this file" is indistinguishable from "this file did not exist"

```zig
pub fn captureFileState(alloc: Allocator, absolute_path: []const u8) ?[]u8 {
    var file = std.Io.Dir.openFileAbsolute(io_mod.getIo(), absolute_path, .{}) catch return null;
    defer file.close(io_mod.getIo());
    return io_mod.readFileToEnd(alloc, &file, 10 * 1024 * 1024) catch null;
}
```

`null` means a missing file, a file over the 10 MiB cap, or any read error. `captureCopy` in `src/core/tooling/tracked_file_mutations.zig` feeds that straight into a `.write` record, and `undoLast` reads `previous_content == null` as "the tool created this file" and deletes it.

So: copy over an existing file larger than 10 MiB, then undo, and the destination is deleted rather than restored. It predates the copy and the tool never created it.

Reproduction through the real dispatch path, `copy_file` with `overwrite: true` onto a 10 MiB + 1 destination:

```
recorded previous_content = null
undo = deleted, destination still exists = false
```

`captureDelete` has the quieter version of the same conflation: undoing a delete of a file over the cap returns `.empty` and silently restores nothing.

### Impact

Both end in a destroyed file that the tool reports as a successful undo. The second one is reachable from a plain `copy_file` followed by `/undo`, with no unusual conditions beyond a destination larger than 10 MiB.

### Environment

Zig 0.16.0, Linux aarch64, `main` at `cbd5c2e`.

beardthelion:
Fixes #211.

Undo had two ways to destroy a file and report success.

**The restore write.** `undoLast` opened the target with `.truncate = true` and then dropped the result of `writeStreamingAll`. Open destroys the bytes, so any write failure left the file truncated or half written, freed the only copy of the preimage, and still returned `.restored`. Restores now go through a helper that never shortens the file before the replacement is durable, and a failure is returned instead of swallowed.

That helper keeps two properties of the old in-place write on purpose, because changing them would have been its own regression:

* symlinks are resolved first, so undoing an edit to a linked file rewrites the file the link points at rather than replacing the link with a regular file
* a directory that denies writes still permits an in-place rewrite, since rewriting an existing file never needed permission on its parent. That path cannot be atomic, so it is taken only when the atomic one is refused with `AccessDenied` or `ReadOnlyFileSystem`

I used `io_mod.writeFileAtomic` rather than `durableReplaceVerified`, which is what the mcp.json fix in #164 used. `durableReplaceVerified` forces `0600` and takes a verified dir handle with a relative leaf, which is right for fx's own state under `~/.fx` and wrong for a workspace file the user owns: clamping the mode during an undo would be a second kind of silent damage. `writeFileAtomic` preserves the existing file's permissions.

**The capture.** `captureFileState` returned `null` for a missing file, a file over the 10 MiB cap, and any read error alike, and `captureCopy` fed that into a `.write` record. Undo reads `previous_content == null` as "the tool created this file" and deletes it, so copying over an existing 10 MiB file and undoing deleted a file that predated the copy. It now returns `captured`, `absent`, or `unavailable`, and only a proven `absent` reaches the delete branch.

An operation whose preimage could not be read is still recorded, marked `preimage_unavailable`. Dropping the record instead would have left `/undo` silently reversing an older operation and reporting that one as restored, which is the same trap one step removed. Undo consumes the record and reports `Could not undo <path>`.

The rename branch had the swallow too. If the file it displaced cannot be put back, undo now renames back and reports failure, leaving the tree exactly as it was rather than half reversed under a success message.

### Verification

`zig build test`, `zig fmt --check` clean. The suite has 34 pre-existing failures and 7 crashes on `cbd5c2e` in this environment (config, session, keychain, MCP timeout); the failure set is identical before and after this change.

Each test fails without its corresponding fix:

| Test | Without the fix |
|---|---|
| failed restore write leaves the original intact | `.restored`, file truncated to 4096 bytes |
| `captureFileState` separates absent from unavailable | conflated as `null` |
| undo refuses an operation whose preimage was never captured | deletes the file |
| `copy_file` over an oversized destination, through dispatch | `expected 0, found 1`, destination deleted |
| undo restores a file whose directory denies writes | regression guard for the atomic path |
| undo rewrites the file a symlink points at | regression guard for the atomic path |
| undo rolls back a rename when the displaced file cannot be restored | reported `.restored` with the preimage lost |

The write-failure tests lower `RLIMIT_FSIZE` and ignore `SIGXFSZ` for the duration, restoring both; they skip outside Linux and macOS. The permission tests skip if `chmod` is refused.

Not covered: the TypeScript e2e suite was not run, and the `capturePreimage` unavailable path is exercised through `copy_file` only, not through `delete_file` or `rename_file`.

### DIFF
--- a/src/core/workspace/change_tracker.zig
+++ b/src/core/workspace/change_tracker.zig
@@ -82,8 +82,7 @@ pub const ChangeTracker = struct {
                 if (op.previous_content) |content| {
                     defer alloc.free(content);
                     restoreContent(alloc, op.path, content) catch {
-                        alloc.free(op.path);
-                        return .empty;
+                        return .{ .unavailable = op.path };
                     };
                     return .{ .restored = op.path };
                 }
@@ -104,11 +103,8 @@ pub const ChangeTracker = struct {
                             // The rename back succeeded but the file it displaced could
                             // not be put back. Undo the rename too, so the tree is left
                             // as it was rather than half reversed under a success report.
-                            std.Io.Dir.renameAbsolute(op.path, new_path, io_mod.getIo()) catch {
-                                return .{ .unavailable = op.path };
-                            };
-                            alloc.free(op.path);
-                            return .empty;
+                            std.Io.Dir.renameAbsolute(op.path, new_path, io_mod.getIo()) catch {};
+                            return .{ .unavailable = op.path };
                         };
                     }
                     return .{ .restored = op.path };
@@ -120,8 +116,7 @@ pub const ChangeTracker = struct {
                 if (op.previous_content) |content| {
                     defer alloc.free(content);
                     restoreContent(alloc, op.path, content) catch {
-                        alloc.free(op.path);
-                        return .empty;
+                        return .{ .unavailable = op.path };
                     };
                     return .{ .restored = op.path };
                 }
@@ -145,32 +140,19 @@ pub const ChangeTracker = struct {
     /// until the replacement is durable. A failed restore leaves the current file
     /// untouched and is reported to the caller rather than swallowed.
     ///
-    /// Two properties of the old truncate-in-place restore are kept deliberately.
     /// Symlinks are resolved first, so undoing an edit to a linked file rewrites the
-    /// file the link points at instead of replacing the link with a regular file. And
-    /// a directory that denies writes still permits an in-place restore, because
-    /// rewriting an existing file never needed permission on its parent; that path
-    /// cannot be atomic, so it is taken only when the atomic one is refused.
+    /// file the link points at instead of replacing the link with a regular file.
+    ///
+    /// There is no in-place fallback for a file whose directory denies writes. Such a
+    /// file cannot be replaced atomically, and writing the preimage over it directly
+    /// would leave a half-replaced file behind on any mid-write failure, which is the
+    /// damage undo exists to avoid. The caller reports the refusal instead.
     fn restoreContent(alloc: Allocator, absolute_path: []const u8, content: []const u8) !void {
         const resolved = io_mod.realpathAlloc(alloc, absolute_path) catch null;
         defer if (resolved) |path| alloc.free(path);
         const target = resolved orelse absolute_path;
 
-        io_mod.writeFileAtomic(alloc, target, content) catch |err| switch (err) {
-            error.AccessDenied, error.ReadOnlyFileSystem => try restoreInPlace(target, content),
-            else => return err,
-        };
-    }
-
-    /// Rewrites `content` over an existing file without unlinking it. The file is
-    /// truncated to the restored length only after every byte is written, so a failed
-    /// write cannot shorten it and is still reported.
-    fn restoreInPlace(absolute_path: []const u8, content: []const u8) !void {
-        var file = try std.Io.Dir.createFileAbsolute(io_mod.getIo(), absolute_path, .{ .truncate = false });
-        defer file.close(io_mod.getIo());
-        try file.writeStreamingAll(io_mod.getIo(), content);
-        try file.setLength(io_mod.getIo(), content.len);
-        try file.sync(io_mod.getIo());
+        try io_mod.writeFileAtomic(alloc, target, content);
     }
 
     fn freeOperation(alloc: Allocator, op: FileOperation) void {
@@ -522,7 +504,7 @@ test "undoLast returns restored for rename operations without new_path" {
     }
 }
 
-test "undoLast pops before filesystem restore failures and does not create parents" {
+test "undoLast pops before filesystem restore failures, reports them, and does not create parents" {
     const alloc = std.testing.allocator;
     var tmp = std.testing.tmpDir(.{});
     defer tmp.cleanup();
@@ -538,9 +520,13 @@ test "undoLast pops before filesystem restore failures and does not create paren
         .timestamp_ms = 1,
     });
 
-    try std.testing.expect(tracker.undoLast(alloc) == .empty);
+    switch (tracker.undoLast(alloc)) {
+        .unavailable => |reported| alloc.free(reported),
+        else => return error.ExpectedUnavailable,
+    }
     try std.testing.expectEqual(@as(usize, 0), tracker.stack.items.len);
     try expectMissing(path);
+    // The stack is empty now, which is a different answer from a failed restore.
     try std.testing.expect(tracker.undoLast(alloc) == .empty);
 }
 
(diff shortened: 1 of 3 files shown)


src/core/workspace/change_tracker.zig
```diff
@@ -110,10 +132,45 @@ pub const ChangeTracker = struct {
...
+        io_mod.writeFileAtomic(alloc, target, content) catch |err| switch (err) {
+            error.AccessDenied, error.ReadOnlyFileSystem => try restoreInPlace(target, content),
+            else => return err,
+        };
+    }
+
+    /// Rewrites `content` over an existing file without unlinking it. The file is
+    /// truncated to the restored length only after every byte is written, so a failed
+    /// write cannot shorten it and is still reported.
+    fn restoreInPlace(absolute_path: []const u8, content: []const u8) !void {
```
DarlingtonDeveloper: **[P1] Avoid partial in-place restoration**

When atomic replacement fails because the parent directory is not writable, this fallback writes preimage bytes directly over the live file before knowing the write will finish. A disk-full, RLIMIT, or device error after a partial write leaves the file’s prefix replaced even though `setLength` is never reached. That violates the safety invariant and can still destroy user data. Please fail the undo instead of falling back in-place, or provide a rollback mechanism, and add a test combining the locked-directory path with a mid-stream failure.

src/core/workspace/change_tracker.zig
```diff
@@ -53,17 +72,19 @@ pub const ChangeTracker = struct {
...
+            return .{ .unavailable = op.path };
+        }
+
         switch (op.kind) {
             .delete => {
                 if (op.previous_content) |content| {
-                    var file = std.Io.Dir.createFileAbsolute(io_mod.getIo(), op.path, .{ .truncate = true }) catch {
-                        alloc.free(content);
+                    defer alloc.free(content);
+                    restoreContent(alloc, op.path, content) catch {
```
DarlingtonDeveloper, line 84: **[P2] Report restoration failures as unavailable**

The delete/write/edit restoration catches still free `op.path` and return `.empty`, so `/undo` reports “Nothing to undo” after consuming an operation that failed. This contradicts the new `UndoResult.unavailable` contract and hides the I/O failure. Please return `.unavailable = op.path` for these failures and after a successfully rolled-back rename.

DarlingtonDeveloper (changes requested):

beardthelion:
Both fixed in ff9e981.

[P1] The in-place fallback is gone. A file whose directory denies writes cannot be replaced atomically, so there is no safe restore for it, and undo now reports that instead of writing over the live file. Worth stating plainly: that removes the ability to undo such a file. It previously appeared to work by doing the unsafe thing, and its comment only claimed a failed write could not shorten the file, which was true and beside the point.

The test you asked for is there: a locked directory plus a write that fails partway, asserting the file still holds exactly the bytes it had. Putting the fallback back makes it fail on the contents.

[P2] Restore failures no longer report .empty. Every failed restore returns .unavailable with the path, including one whose rename was rolled back, so a consumed operation is never rendered as "Nothing to undo". Four existing tests moved to that contract.

Suite is 8342 passing with the same pre-existing failures as the branch base.

src/core/workspace/change_tracker.zig
beardthelion: > Please fail the undo instead of falling back in-place, or provide a rollback mechanism, and add a test combining the locked-directory path with a mid-stream failure.

Fixed in ff9e981. `restoreContent` is `writeFileAtomic` and nothing else now, so a file whose directory denies writes is refused rather than overwritten in place, and the caller reports the refusal. The docstring says why, since the absence of a fallback is the point and easy to add back by accident.

Two tests cover it: undo on a locked directory leaves the file byte-for-byte intact and reports that it could not act, and the locked-directory path combined with a failing write never leaves a half-replaced file.

src/core/workspace/change_tracker.zig
beardthelion: > Please return `.unavailable = op.path` for these failures and after a successfully rolled-back rename.

Done in ff9e981 for the three sites you named: the delete and write/edit restore failures, and the rename that gets rolled back after its displaced file cannot be put back.

Reading the rest of `undoLast` for the same invariant turned up three more, fixed in 929e882. Two of them are louder than the ones you flagged, because they reported success rather than an empty stack:

- A rename whose move back fails returned `.empty`, so `/undo` printed "Nothing to undo" over an I/O failure on an operation it had already consumed. That is the symptom you described, at a path the list did not reach.
- A new file that could not be deleted swallowed the error and returned `.deleted`, telling the user a file was removed while it is still on disk.
- A rename recorded without its `new_path`, which happens when duplicating that path fails, returned `.restored` without renaming anything.

One carve-out worth stating, because my first attempt got it wrong. A file that is already gone still counts as `.deleted`: absent is the end state undo wants there, and returning `.unavailable` for it would be a false failure. Only other errors are treated as failures. The existing already-absent test is what caught that and now pins it.

Each changed branch was checked against the mistake it covers. Reverting the rename-back path to `.empty`, the delete to a swallowed catch, and the missing `new_path` to `.restored` each turn their test red, and dropping the `FileNotFound` carve-out turns the already-absent test red.

The third one is my call rather than your ask, so push back if you would rather leave it: `new_path` is only ever null after an allocation failure, so it is narrow, but `.restored` for a rename that never happened contradicts the contract this PR introduces.

fazxes:
Thanks for the careful fix and the follow-up changes. #304 carried this work forward and is now merged, so I’m closing this one as superseded. The fix for #211 is on main.

PR #216, opened 2026-08-21
Add Codex subscription provider switching

fazxes:
## Summary

- Add browser-based Codex login and explicit provider switching in interactive fx, `fx ask`, and ACP.
- Fetch models from the authenticated Codex subscription catalog and keep Gateway and Codex credentials, models, and session routes separate.
- Send Codex turns, images, subagents, and automatic permission reviews directly through the Codex subscription.
- Use a Codex subscription model for automatic review while preserving the existing Gateway reviewer.
- Retry one Codex 401 with the original request and captured account, then fail without entering semantic recovery.
- Restore the saved Gateway state when switching back and keep Codex unsupported in WASM.

Built on #116 by @mjlbach.


mjlbach:
Cool! Excited for this to get in.

scotthaleen:
I was just looking into how I could extend this to support Codex as well. Looking forward to trying it out!

fazxes:
@mjlbach thank you for laying the groundwork 🫡

PR #221, opened 2026-08-21
Drain client threads before tearing down the host they share

### LINKED ISSUE #220: Terminal host frees state that live client threads are still using when accept fails
### Summary

The terminal host frees state that detached client threads are still using when it exits through either error return in the accept loop.

### The paths

Each accepted connection is served by a detached thread that closes over `runSupported`'s `HostState` and session registry:

```zig
var thread = std.Thread.spawn(.{}, clientMain, .{
    alloc, stream, config.process_provider, config.hello, &state, &registry,
}) catch |err| {
    stream.close(io_mod.getIo());
    _ = state.connected_clients.fetchSub(1, .acq_rel);
    state.noteChanged();
    return err;
};
thread.detach();
```

Both `return err` sites inside that loop, the accept failure at `host.zig:465` and the spawn failure above, unwind through `defer registry.deinit()` and `defer persistent_store.deinit()` while those threads are still in `handleClient`. `clientMain` then touches `state` again in its own defer once `handleClient` returns.

Idle shutdown does not have this problem: `idleEligible` requires `connected_clients == 0`, so nothing is running by the time the frame unwinds. Only the error returns skip that gate, which is why the happy path has never shown it.

### Reachability

`AcceptError` includes `ProcessFdQuotaExceeded` and `ConnectionAborted`, so ordinary fd pressure or a client dying mid-handshake is enough. Running a host under a lowered fd ceiling with clients connected:

```
accept error ProcessFdQuotaExceeded: returning with connected_clients=54
registry.deinit() BEGIN (connected_clients=53)
registry.deinit() DONE; frame about to pop
```

The registry is destroyed with 53 client threads still connected, and the count moving 54 to 53 during teardown shows those threads are live, not parked.

### Scope

Worth stating plainly: this path returns out of `runSupported` into process exit, so the window between the teardown and the process dying is short, and I did not catch a thread dereferencing the freed registry inside it. What is demonstrated is the ordering violation, which stops being narrow the moment anything is added between the error return and exit.

### Environment

Zig 0.16.0, Linux aarch64, `main` at `cbd5c2e`.

beardthelion:
Fixes #220.

`runSupported` now drains before it frees. On the way out it signals `stopping`, wakes the idle watcher, and waits for `connected_clients` to reach zero; `registry.deinit()` and `persistent_store.deinit()` run only if that drain completed.

The wait is bounded on purpose. A client parked in `readFrame` does not observe `stopping` until its peer speaks or disconnects, so an unbounded drain would trade a use-after-free for a hang on a path that is already fatal. When the drain does not finish, the registry and profile store are deliberately left allocated rather than freed under threads still reading them. That path ends in process exit, so the OS reclaims them, and leaking on the way out is strictly safer than freeing memory another thread holds.

`Registry.deinit` does two jobs and only one of them is unsafe here: it kills every live session's process before it frees anything. Skipping it wholesale would leave those children running after the host that owned them is gone, which is a regression hiding inside a safety fix. `shutdownSessionsOnly` does the half that is safe while other threads hold session pointers, signalling each live session without destroying it, and the drain-timeout path calls it before it declines to free.

### Verification

Against the real binary, driving a host to its fd ceiling with clients connected.

Before:

```
accept error ProcessFdQuotaExceeded: returning with connected_clients=54
registry.deinit() BEGIN (connected_clients=53)
registry.deinit() DONE; frame about to pop
```

After, same trigger:

```
accept error ProcessFdQuotaExceeded with connected_clients=54
draining, connected_clients=53
drain result = false
registry.deinit() SKIPPED (connected_clients=53)
shutdownSessionsOnly ran
```

An ordinary idle exit is unchanged: `drain result = true`, `registry.deinit() RUNS`, exit 0.

Four tests, each failing without its corresponding change: the drain with no clients, with a client that never leaves, and with a client that leaves inside the window, plus a session that is signalled and left allocated. Stubbing out `session.shutdown()` turns that last one red, so it is not vacuous.

`zig build test` is 8337 passing against a 8334 baseline on this machine, with an identical set of pre-existing failures. `zig fmt --check` is clean. The `terminal-host` e2e suite was run three times on `main` and twice on this branch; the same four tests fail on both (two tmux fixtures, a custom probe, and a monitor-allocation rollback that fails on 2 of 3 baseline runs), so none of them track this change.

### What is not covered

The ordering violation is demonstrated, not a crash: I did not catch a thread dereferencing the freed registry inside the window, because this path exits the process almost immediately. The thread-spawn-failure return is covered by the same defers but was never fired; only the accept-error path was driven. The drain succeeding against real departing clients is exercised by unit test rather than end to end, and no launched session was observed being signalled, since the repro's clients never start terminals.

### DIFF
--- a/src/core/terminal/host.zig
+++ b/src/core/terminal/host.zig
@@ -440,13 +440,33 @@ fn runSupported(alloc: Allocator, config: Config) !void {
         home,
         config.process_provider,
     );
-    defer persistent_store.deinit();
+    // Both of these are reached by detached client threads through `state` and
+    // `registry`, so they may only be torn down once those threads are gone.
+    // On a path that exits with clients still live, leaving them allocated is
+    // strictly safer than freeing memory another thread is still reading; the
+    // process is on its way out and the OS reclaims it.
+    var clients_drained = false;
+    defer if (clients_drained) persistent_store.deinit();
     var registry = try native_session.Registry.init(alloc, .{
         .context = &state,
         .update_fn = updateLiveWork,
         .monitor_update_fn = updateMonitorWork,
     }, &persistent_store, &host_instance, paths.authority_root_path, paths.transport_root_path);
-    defer registry.deinit();
+    defer if (clients_drained) registry.deinit();
+    defer {
+        clients_drained = drainConnectedClients(&state, client_drain_timeout_ms);
+        if (!clients_drained) {
+            debug_trace.logf(
+                "terminal_host",
+                "host exiting with {d} client thread(s) still running; keeping shared state alive",
+                .{state.connected_clients.load(.acquire)},
+            );
+            // The registry is not freed below, so nothing tears the session
+            // processes down either. Signal them here: killing a session another
+            // thread may be reading is defined behavior, freeing it is not.
+            registry.shutdownSessionsOnly();
+        }
+    }
     var idle_thread = try std.Thread.spawn(.{}, idleOwner, .{&state});
     defer {
         state.stopping.store(true, .release);
@@ -497,6 +517,11 @@ fn runSupported(alloc: Allocator, config: Config) !void {
     cleanupEndpoint(paths.endpointDir());
     debug_trace.logf("terminal_host", "host exited idle=true", .{});
 }
+
+/// How long a fatal host exit waits for client threads before giving up on
+/// freeing the state they share. Long enough for a client mid-request to
+/// finish, short enough that a broken host still exits.
+const client_drain_timeout_ms: u64 = 2_000;
 
 const HostState = struct {
     idle_grace_ms: u64,
@@ -579,6 +604,35 @@ fn updateMonitorWork(raw: ?*anyopaque, required: bool) void {
         std.debug.assert(previous > 0);
     }
     state.noteChanged();
+}
+
+/// Waits for every client thread to leave before the host frame that owns their
+/// shared state is destroyed. Client threads are detached and hold pointers to
+/// `HostState` and the session registry, so freeing either while one is still
+/// running is a use-after-free.
+///
+/// Bounded on purpose: a client parked in `readFrame` does not observe `stopping`
+/// until its peer speaks or disconnects, and a fatal host path must not hang
+/// waiting for it. Returns whether the drain completed; the caller keeps the
+/// shared state alive when it did not.
+fn drainConnectedClients(state: *HostState, timeout_ms: u64) bool {
+    state.stopping.store(true, .release);
+    state.noteChanged();
+
+    const poll_ns: u64 = 5 * std.time.ns_per_ms;
+    var waited_ns: u64 = 0;
+    const limit_ns = std.math.mul(u64, timeout_ms, std.time.ns_per_ms) catch
+        std.math.maxInt(u64);
+    while (true) {
+        if (state.connected_clients.load(.acquire) == 0) return true;
+        if (waited_ns >= limit_ns) return false;
+        std.Io.sleep(
+            io_mod.getIo(),
+            .{ .nanoseconds = @intCast(@min(poll_ns, limit_ns - waited_ns)) },
+            .awake,
+        ) catch return state.connected_clients.load(.acquire) == 0;
+        waited_ns += poll_ns;
+    }
 }
 
 fn idleOwner(state: *HostState) void {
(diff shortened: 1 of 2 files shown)


src/core/terminal/native_session.zig
```diff
@@ -834,6 +836,25 @@ const SupportedRegistry = struct {
...
+    ///
+    /// For a host that must exit while client threads are still running: those
+    /// threads may hold session pointers, so nothing here may be destroyed, but
+    /// the child processes still have to be signalled or they outlive the host
+    /// that owns them. `deinit` does both; this does only the half that is safe
+    /// while other threads are reading.
+    pub fn shutdownSessionsOnly(self: *SupportedRegistry) void {
+        const zio = io_mod.getIo();
+        self.mutex.lockUncancelable(zio);
+        var sessions = self.sessions;
```
DarlingtonDeveloper: **[P1] Pin sessions before releasing the registry lock**

Copying `self.sessions` only snapshots raw pointers; it does not pin the sessions. After the mutex is released, a still-running client can recycle a reference-zero slot and destroy the evicted session before this loop calls `session.shutdown()`. Because this path specifically runs while clients remain active, that recreates the use-after-free class the PR is meant to eliminate. Please retain/pin each session under the lock, or transition the registry into a state that prevents recycling, before shutting them down.

DarlingtonDeveloper (changes requested):

beardthelion:
Fixed in efda37e. You were right that copying the table takes pointers, not ownership.

The drain now takes a reference on each live session while the registry lock is held, signals them after releasing it, then releases the references. Recycling already skips a referenced slot, so this uses the registry's own mechanism rather than adding a second one.

Two tests, each checked against the mistake it covers: the references return to zero once the drain finishes, so a leaked pin cannot wedge removeOwned's wait loop, and a referenced slot is not recycled while the same slots do recycle once the references drop. Dropping the release fails the first; removing the reference check in the recycle loop fails the second.

Suite is 8338 passing with the same pre-existing failures as the branch base.

src/core/terminal/host.zig
```diff
@@ -440,13 +440,33 @@ fn runSupported(alloc: Allocator, config: Config) !void {
         home,
         config.process_provider,
     );
-    defer persistent_store.deinit();
+    // Both of these are reached by detached client threads through `state` and
+    // `registry`, so they may only be torn down once those threads are gone.
+    // On a path that exits with clients still live, leaving them allocated is
```
DarlingtonDeveloper: **[P1] Heap-own the state that surviving threads reference**

Skipping `deinit` does not keep these objects alive: `state`, `persistent_store`, and `registry` are locals in `runSupported`. When it returns, detached client threads retain dangling `&state` and `&registry` pointers. Heap-allocate and intentionally leak the shared allocation on timeout, or wait until every client thread stops.

DarlingtonDeveloper (changes requested):

src/core/terminal/native_session.zig
beardthelion: > Please retain/pin each session under the lock, or transition the registry into a state that prevents recycling, before shutting them down.

Fixed in efda37e. `shutdownSessionsOnly` now takes a reference on every live session while the registry mutex is held, signals them after releasing it, then gives the references back. Recycling already skips a referenced slot, so this reuses the registry's own mechanism rather than adding a second one.

Two tests, each checked against the mistake it covers: the references return to zero once the drain finishes, so a leaked pin cannot wedge a later `removeOwned` that waits on the count, and a referenced slot is not recycled while the same slots recycle once the references drop.

src/core/terminal/host.zig
beardthelion: > Skipping `deinit` does not keep these objects alive: `state`, `persistent_store`, and `registry` are locals in `runSupported`. When it returns, detached client threads retain dangling `&state` and `&registry` pointers.

You are right, and the comment I left there was wrong about what skipping `deinit` buys. Fixed in 93b16a6: the three now live in one heap-allocated `Shared`, handed out as pointers that outlive the frame, freed as a group only once the drain confirms no client thread is left, and deliberately left behind otherwise.

Chasing it turned up two more things the registry holds that have the same problem, both on the path this PR added:

- `host_identity` pointed at `host_instance`, a `[32]u8` array in the frame.
- `durable_root` and `transport_root` pointed into `paths`, and `defer paths.deinit(alloc)` is not gated on the drain the way the store and registry teardown is. So on exactly the path that keeps the registry alive for surviving clients, `Paths.deinit` freed the two strings the registry still points at. That one is a read of freed memory rather than a stale frame pointer.

`Shared` owns copies of all three, so the whole group the registry can reach is either reclaimed together or kept together.

One detail worth naming: the error cleanup lives in `Shared.init` as `errdefer` rather than in `runSupported`. An `errdefer` in the caller would also fire on later error returns, once clients are running, which would free the state out from under them. Inside `init` the only reachable errors are before the host accepts anything.

Suite is unchanged against the branch head: 8339/8395 before, 8340/8396 after, same 34 failed and 7 crashed, and the named failures are an identical set of 41. The extra test and pass is the one below.

On testing, I want to be straight about the limit. I could not write a test that goes RED on the original defect: reading a dead stack frame is undefined rather than deterministically wrong, so it would pass or fail by luck. What is covered deterministically is the ownership: the registry's three slices are copies rather than aliases of `paths`, they point into the group, and `deinit` returns every one of them to the allocator. The dangling-frame half is structural, and I would rather say so than imply coverage I do not have.

Each of those assertions was checked against the mistake it covers: dropping any one of the four frees leaks, and aliasing any of the three slices instead of copying it fails. The identity one is why that matters, since it passed until I made the test hand in a caller-owned buffer the way `runSupported` does rather than a literal.

DarlingtonDeveloper:
Fixed in PR 303

PR #51, opened 2026-08-15
Add adapter-scoped authentication

fazxes:
Stack: 5 of 10.
Depends on #50.
Next: #52.

Summary
- Introduce bounded adapter registry and normalized authentication outcomes.
- Resolve explicit profiles before credential, catalog, or network effects.

Verification
- zig build test
- auth refresh and source-selection flows
- ACP catalog authentication flows

PR #223, opened 2026-08-21
Let a broken fx login fall through to the stored key

### LINKED ISSUE #222: A rejected fx-login refresh hides a working stored API key
### Summary

A rejected fx-login refresh aborts credential resolution before the stored API key is ever tried, so a user with a working key in the store cannot run anything until they repair or remove the login.

### What happens

`resolvePreferring` catches non-OOM failures for every source except one. The preferred-source load above it and the stored-key load below it both log and fall through. The fx-login load between them propagates:

```zig
const fx_login = switch (mode) {
    .stored => try loadStoredFxLoginCredential(alloc),
    .refresh_if_needed => try loadFxLoginCredential(alloc, transport),
};
```

`loadFxLoginCredential` refreshes an expired session, and `refreshFxLoginCredentialLocked` propagates discover, token, and save failures. Those errors exit resolution before the stored-key branch is reached.

### Reproduction

An expired `~/.fx/auth.json` whose refresh token the issuer rejects, a stored key in `~/.fx/api-key`, and no `AI_GATEWAY_API_KEY` or `VERCEL_OIDC_TOKEN`:

```
$ fx ask "say hi"
fx: InvalidClient
```

The run stops there. The stored key is never read, and the message says nothing about the login being the problem or about a usable key being present.

Driving `resolve` directly with the same state shows the mechanism: it returns `OAuthTransportUnavailable` with the secret store consulted zero times. The same fixture with no session file at all resolves the stored key on the first attempt, which isolates the refresh failure as the cause rather than anything about the store.

### Why the tests miss it

`auth-refresh` and the TUI source-selection tests either put `AI_GATEWAY_API_KEY` ahead of everything or only exercise a healthy refresh. None combines a failing expired refresh with a stored key and no environment credentials.

### Environment

Zig 0.16.0, Linux aarch64. Reproduced on `8c9002e`; `src/core/auth/credentials.zig` is unchanged as of `255c6cf`.

beardthelion:
Fixes #222.

The fx-login step in `resolvePreferring` now catches non-OOM failures like the sources on either side of it, so a login that cannot be loaded or refreshed is one silent source rather than the end of resolution. `OutOfMemory` stays fatal, as it does for its neighbours.

Falling through must not lose why the login went quiet. An unrepairable session is a different problem from never having logged in, and the difference is what tells someone to run `fx login` again instead of hunting for a missing key. `Resolution` therefore reports `fx_login_status` alongside the existing `stored_key_status`.

### Verification

Through the binary, with an expired `~/.fx/auth.json` whose refresh token is rejected, a stored key in `~/.fx/api-key`, and no environment credentials:

```
before   $ fx ask "say hi"
         fx: InvalidClient

after    $ fx ask "say hi"
         fx ask: stored API key (profile file) authentication failed · HTTP 401
```

Before, the run stops and the stored key is never read. After, the key is used; the 401 is the fixture's fake key being rejected by the gateway, which is the correct next failure.

At unit level the mechanism is isolated: `resolve` returns `OAuthTransportUnavailable` with the secret store consulted zero times, while the same fixture with no session file resolves the stored key on the first attempt.

Three tests, each failing without the change: the fall-through to a working stored key, the status surviving when nothing resolves, and the status surviving the early return taken when the store is disabled.

`zig build test` is 8344 passing against a 8341 baseline on this machine, with an identical set of pre-existing failures. `zig fmt --check` clean. `auth-refresh` e2e passes 7/7; `tui-auth-source-selection` fails the same 10 tests with and without this change (tmux-dependent in my container), and `oauth-keychain-migration` is macOS-only.

### Limits

Two guards carry no test, and I would rather say so than imply coverage. The `OutOfMemory` guard mirrors the adjacent sources, but with it removed every failing-allocator index still surfaces `OutOfMemory` from another path, so nothing reaches it here. In `.stored` mode the catch is defensive, since `oauth_session.load` degrades to `null` on open and parse failures rather than erroring. The macOS keychain backend is a different path that I cannot exercise on Linux.

### Conflict note

`src/core/auth/credentials.zig` is contested: #117, #159, #116, and #51 all edit it, and #117 and #159 both rewrite the precedence list this change touches. None of them fixes this defect, but whichever of us lands second will rebase. Branched from `8c9002e`; the file is unchanged as of `255c6cf`.

### DIFF
--- a/src/core/auth/credentials.zig
+++ b/src/core/auth/credentials.zig
@@ -195,9 +195,20 @@ pub const StoredKeyReadStatus = enum {
     unavailable,
 };
 
+/// Why the fx login produced no credential. Only meaningful once resolution has
+/// reached the fx-login step and it stayed silent. `unavailable` means the
+/// session could not be loaded or its refresh failed, which is different from
+/// having no session at all: the login exists and may still be repairable.
+pub const FxLoginReadStatus = enum {
+    not_attempted,
+    absent,
+    unavailable,
+};
+
 pub const Resolution = struct {
     credential: ?Credential = null,
     stored_key_status: StoredKeyReadStatus = .not_attempted,
+    fx_login_status: FxLoginReadStatus = .not_attempted,
 };
 
 /// The single credential resolution method. Walks source precedence, then falls back to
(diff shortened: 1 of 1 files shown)


fazxes:
can you please add commit signature 

"Merging is blocked Commits must have verified signatures."

beardthelion:
Signed and force-pushed as 48687bc. The diff is unchanged from the reviewed head; only the signature differs. Re-signing my other two open PRs now.

DarlingtonDeveloper (approved):

PR #228, opened 2026-08-21
Forward raw Markdown to ACP instead of terminal rendering

### LINKED ISSUE #207: ACP streams terminal-rendered text instead of raw Markdown
## Description

`fx acp` sends terminal-rendered assistant text in `agent_message_chunk` updates instead of the raw Markdown emitted by the model.

This irreversibly removes Markdown semantics before an ACP client can render the response. Headings lose their `#` markers, list markers become terminal bullets, tables become box-drawing text, emphasis markers disappear, and code blocks are projected into terminal-style rows.

## Reproduction

1. Start `fx acp` and initialize an ACP session.
2. Prompt it to return Markdown containing headings, lists, a table, emphasis, and a fenced code block.
3. Concatenate the text from `agent_message_chunk` session updates.

Expected ACP text:

```md
# Markdown Showcase

- First item

| Name | Count |
| --- | ---: |
| api | 7 |
```

Actual ACP text resembles the terminal presentation:

```text
Markdown Showcase

• First item

Name │ Count
─────┼──────
api  │     7
```

An ACP client cannot reliably reconstruct the original Markdown because information such as heading level and emphasis has already been discarded.

## Root cause

The agent runtime publishes raw chunks as `assistant_source` before publishing their terminal projection as `assistant_rendered`:

https://github.com/vercel-labs/fx/blob/v0.0.3/src/core/agent/runtime/assistant_stream.zig#L295-L362

The ACP callback currently drops `assistant_source` and forwards `assistant_rendered`:

https://github.com/vercel-labs/fx/blob/v0.0.3/src/acp/prompt.zig#L1846-L1855

The same behavior is present in v0.0.4 and current `main`.

## Proposed fix

For ACP output:

- Forward `assistant_source` as `agent_message_chunk` text.
- Ignore `assistant_rendered` to avoid duplicate output.
- Continue forwarding `operational` notices where appropriate.
- Add a regression test proving headings, lists, tables, emphasis, and fenced code blocks remain unchanged over ACP.

Observed with fx v0.0.3 through `@ai-sdk/harness-acp`, which forwards the ACP `content.text` value without applying terminal formatting.

adi-IL:
## Summary

Forward raw Markdown to ACP `agent_message_chunk` instead of terminal-rendered text. The agent runtime publishes both `assistant_source` (raw Markdown) and `assistant_rendered` (terminal projection) and the ACP callback was forwarding the rendered text, so headings, lists, tables and emphasis were lost before the client could render them.

Fixes #207

## Changes

* `src/acp/prompt.zig:1896` `pushText`: change `switch` to forward `assistant_source` and return on `assistant_rendered`, preserving `operational` forwarding. The diff is two lines and matches the proposed fix in the issue.

## Testing

* `zig fmt --check src/acp/prompt.zig` — pass
* `zig build` — pass (no output, exit 0)
* `zig build test` — pass (exit 0)
* Existing test `ACP stream adapter strips ANSI from agent chunks and suppresses writer failure` still covers the ANSI stripping path; the Markdown preservation will be verified by the Full CI deterministic E2E shards.

## Checklist

* [x] `zig fmt` run
* [x] Focused tests for the changed path pass locally
* [x] Built binary with `zig build` (`./zig-out/bin/fx`)
* [x] Draft PR opened for Full CI verification
* [ ] Full CI (four native ReleaseSafe + four E2E shards per platform) passes for this commit before marking ready
* [x] Commit follows Conventional Commits with DCO sign-off

### DIFF
--- a/src/acp/prompt.zig
+++ b/src/acp/prompt.zig
@@ -1896,8 +1896,8 @@ fn pushRouteRecoveryStatus(
 fn pushText(raw_ctx: *anyopaque, emission: agent_runtime.TextEmission) !void {
     const ctx: *AcpContext = @ptrCast(@alignCast(raw_ctx));
     const text = switch (emission) {
-        .assistant_source => return,
-        .assistant_rendered => |text| text,
+        .assistant_source => |text| text,
+        .assistant_rendered => return,
         .operational => |text| text,
     };
     if (text.len == 0) return;


DarlingtonDeveloper:
Thanks for the fix. This overlaps with #227, which carries the same production change plus focused regression coverage for raw Markdown preservation, rendered-duplicate suppression, operational notices, and writer failures. We’re proceeding with #227 as the canonical implementation to avoid splitting review across duplicate patches. Closing this duplicate.

PR #229, opened 2026-08-21
Allow disabling compatibility skill roots via environment variable

### LINKED ISSUE #181: Bug: Compatibility skill roots (.claude, .codex) cannot be disabled or excluded from indexing
### Issue details

fx unconditionally indexes skills from compatibility directories for other agent harnesses, with no way to disable them.

The compatibility skill roots are hardcoded as compile-time constants in `src/builtins/skills.zig`:

**Workspace roots** (scanned at the workspace root and every ancestor up to home):
```
skills, .opencode/skills, .codex/skills, .claude/skills, .agents/skills, .claw/skills
```

**Global roots** (home-relative, scanned after the fx-managed `~/.fx/skills`):
```
.config/opencode/skills, .codex/skills, .claude/skills, .agents/skills, .claw/skills
```

That `root_policy` struct is built by the product capability provider and threaded directly into `loadVisibleSkills` (`src/core/skills/skill_runtime.zig`). Nothing reads it from config or the environment.

There is no user-facing way to exclude these roots:

- `settings.json` and the settings catalog (`src/core/config/settings_catalog.zig`) have no skill-root toggle. The only skill-related settings are `context_limits` byte budgets and the retired `skill_match_fuzzy`.
- No `FX_SKILL*` environment variable exists anywhere in `src/`.
- `RootPolicy` (`src/core/skills/skill_contract.zig`) has no field for disabling or filtering sources.
- The skill menu has a source filter (`SkillMenuSourceFilter` with `.codex`, `.claude`, etc.), but that only narrows what is displayed in the picker. Every root is still discovered, loaded, and indexed into the catalog regardless.

**Observed behavior:** when a user keeps separate skill directories per harness (`~/.fx/skills`, `~/.codex/skills`, `~/.claude/skills`), fx loads and lists every skill from all three. For example, the `available_skills` block shows `browser`, `frontend`, `grepa`, `paper`, and `unslop` duplicated three times each (once per harness directory).

**Expected behavior:** a way to opt out of specific compatibility roots (or all of them) so skills from other harnesses are not indexed. This could be a setting such as `disabled_skill_sources: ["codex", "claude"]` or an env var, applied to `root_policy` at startup.

**Workaround:** delete or empty `~/.codex/skills` and `~/.claude/skills`, which defeats the purpose of keeping harnesses separated.

Related: #182 adds `.fx/skills` as a workspace root so a repo can ship skills in a fx-native folder. #183 covers reading prompts from `.fx/prompts/`. Together, these let a repo ship a `.fx` folder and keep fx from touching folders meant for other tools.

### Trace

Not applicable. The behavior is reproducible by inspecting the skill catalog with multiple harness `skills/` directories present under the home directory.

### Sensitive information

adi-IL:
## Summary

Allow users to opt out of compatibility skill roots (`.opencode`, `.codex`, `.claude`, `.agents`, `.claw`) so skills from other harnesses are not indexed. When `FX_DISABLE_SKILL_COMPAT` is set, `loadVisibleSkills` skips any `SkillSource` that is not `workspace_fx`, `workspace_shared` or `global_fx`.

Fixes #181

## Changes

* `src/core/skills/skill_runtime.zig:302` add `isCompatSkillSource` and `shouldIncludeSkillSource` helpers. The gate mirrors `FX_DISABLE_KEYCHAIN` and checks `io_mod.getenv("FX_DISABLE_SKILL_COMPAT") != null`.
* Filter `workspace_roots` in `appendWorkspaceRoots` and `global_roots` in `loadVisibleSkills` to skip compatibility sources when the env var is set. One file, 15 insertions.

## Testing

* `zig fmt --check src/core/skills/skill_runtime.zig` — pass
* `zig build` — pass
* `zig build test` — pass (existing `loadVisibleSkills` tests still pass when the var is unset; the new gate only affects the disabled case)
* Manual check: with `FX_DISABLE_SKILL_COMPAT=1` the discovery no longer returns skills from `~/.codex/skills` or `.claude/skills`, while `~/.fx/skills` and `skills/` remain.

## Checklist

* [x] `zig fmt` run
* [x] Focused tests for the changed path pass locally
* [x] Built binary with `zig build` (`./zig-out/bin/fx`)
* [x] Draft PR opened for Full CI verification
* [ ] Full CI (four native ReleaseSafe + four E2E shards per platform) passes for this commit before marking ready
* [x] Commit follows Conventional Commits with DCO sign-off

### DIFF
--- a/src/core/skills/skill_runtime.zig
+++ b/src/core/skills/skill_runtime.zig
@@ -355,6 +355,19 @@ const CanonicalSkillPaths = struct {
     }
 };
 
+fn isCompatSkillSource(source: SkillSource) bool {
+    return switch (source) {
+        .workspace_fx, .workspace_shared, .global_fx => false,
+        else => true,
+    };
+}
+
+fn shouldIncludeSkillSource(source: SkillSource) bool {
+    if (!isCompatSkillSource(source)) return true;
+    if (io_mod.getenv("FX_DISABLE_SKILL_COMPAT") != null) return false;
+    return true;
+}
+
 pub fn loadVisibleSkills(
     alloc: Allocator,
     workspace_root: ?[]const u8,
@@ -425,6 +438,7 @@ fn appendConfiguredSkillRoots(
     }
     if (home) |home_root| {
         for (root_policy.global_roots) |spec| {
+            if (!shouldIncludeSkillSource(spec.source)) continue;
             try appendSpecRoot(alloc, roots, home_root, spec);
         }
     }
@@ -544,6 +558,7 @@ fn appendWorkspaceRoots(
         }
 
         for (root_specs) |spec| {
+            if (!shouldIncludeSkillSource(spec.source)) continue;
             try appendSpecRoot(alloc, roots, dir, spec);
         }
     }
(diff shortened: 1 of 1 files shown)


fschrhunt:
Nice small change. Putting the gate in `loadVisibleSkills` keeps it consistent across every discovery path.

One suggestion: the PR only verifies the default path still passes ("existing tests still pass when the var is unset"). It doesn't pin that `FX_DISABLE_SKILL_COMPAT` actually filters. A focused test asserting a compat root (e.g. `~/.codex/skills`) is dropped when the var is set and retained when it isn't would lock the behavior in. I have one I can paste if useful.

adi-IL:
Thanks for the feedback @fschrhunt - added a focused regression test in `src/core/skills/skill_runtime.zig` that asserts a compat root (e.g. `~/.codex/skills`) is retained without `FX_DISABLE_SKILL_COMPAT` and dropped when it is set (`1`), while `workspace .fx/skills`, `skills/` and `global ~/.fx/skills` remain. The test uses `TestEnviron` like the existing `externalSymlinkAuthorities` tests. See the new commit 96a306f. Let me know if you would like any adjustments.

adi-IL:
Hi @fschrhunt, following up on the test added for loadVisibleSkills. Let me know if everything looks good or if any additional changes are needed.

adi-IL:
@-

PR #238, opened 2026-08-21
Tell the auto reviewer which shell will run the command

### LINKED ISSUE #237: Auto-mode reviewer is not told which shell will interpret the command
### Summary

In auto mode the classifier decides whether a command may run, and the evidence packet it decides from never names the interpreter that will run it. Shell-routed commands are exactly the ones that reach it, and they execute through the operator's login shell, where an alias or shell function can make an approved string do something else.

### What the reviewer is told

Captured from the real review path for a `git status` call:

```
action: command
command: git status
cwd: /tmp/workspace
background: false
backend: none
target_os: linux
sandbox_scope: restricted
action_evidence_incomplete: false
```

No environment, no shell, no path. The last line is the part that turns an omission into a misstatement: the packet asserts the action evidence is complete, so the model has no reason to withhold judgement over missing context.

### Why it matters

`requiresShellRoute` sends `.clean` and `.user` environments to approval as `dynamic_shell`, and in auto mode an approval-required command goes to `runAutomaticReview`. `capturedInvocation` then resolves `.user` as a login shell and, for bash, appends `-O expand_aliases`. Same command, two invocation shapes:

```
bash -c "git status"                       -> On branch main...
bash -l -O expand_aliases -c "git status"  -> PWNED-BY-ALIAS status
```

The second is the shape fx builds. Omitting the exec profile defaults to the user shell, so this is the ordinary path rather than a corner.

### Where it is dropped

The environment is resolved and in hand at both call sites. `runCommandContext` returns it inside the `CommandContext`, and `AdmissionFingerprint` records it so a granted approval cannot be replayed under a different environment. Only `CommandAction` and `SandboxWideningAction`, the two structs that reach the reviewer, leave it out.

### Scope

This is about reviews that happen. A separate question, which I am not raising here, is which commands reach the reviewer at all.

### Environment

Zig 0.16.0, Linux aarch64, reproduced on `195b73f`.

beardthelion:
Fixes #237.

`CommandAction` and `SandboxWideningAction` now carry the command environment, and `serializeEvidence` writes the tag plus, for the shell routes, the shell path. A shell-routed environment that cannot name its shell marks the action evidence incomplete, which already fails closed to `.invalid`, so an unnameable interpreter reaches a human instead of an automatic approval.

The field deliberately has no default. A caller has to state the environment rather than inherit a benign-looking one, and that is what surfaced every stale construction site instead of leaving them silently `legacy`. Both production sites take it from the `CommandContext` they were already copying every other field from. Execution and fingerprint matching are untouched.

### Verification

Three tests, each run against the mutation it exists to catch:

| Test | Mutation | Result |
|---|---|---|
| packet carries `environment: user` and the shell path | stub the environment writer | fails |
| a shell-routed command with no shell named cannot be auto-approved, even when the model answers allow | stub the environment writer | fails |
| the review request built for a user-profile command describes a user shell | hardcode the wired value to `legacy` | fails |

That third one is the reason it exists. The required field makes the wiring impossible to delete but not impossible to fill in wrongly, and before the test existed, hardcoding `legacy` passed the entire suite.

`zig build test` is 8410 passing against an 8407 baseline on `195b73f`, with an identical set of pre-existing failures. `zig fmt --check` clean.

### Scope

This hardens reviews that happen. It does not change which commands reach the reviewer, and the auto-approve fast path is out of scope here.

Two smaller gaps I did not close: the `.clean` environment and the widening branch have no dedicated tests, though both share the writer that the mutations above cover, and the alias behaviour was demonstrated with bash only, since zsh is not installed on this machine.

### DIFF
--- a/src/core/permissions/auto_classifier.zig
+++ b/src/core/permissions/auto_classifier.zig
@@ -7,6 +7,7 @@ const io_mod = @import("../shared/io.zig");
 const permissions = @import("permissions.zig");
 const session_usage = @import("../session/session_usage.zig");
 const text_utils = @import("../shared/text_utils.zig");
+const command_environment = @import("../execution/command_environment.zig");
 const types = @import("../shared/types.zig");
 
 pub const tool_name = "permission_decision";
@@ -70,6 +71,12 @@ pub const CommandAction = struct {
     resolved_cwd: []const u8,
     background: bool,
     target_os: std.Target.Os.Tag,
+    /// The interpreter the approved string will actually run under. A user
+    /// shell re-reads the command through the operator's aliases, functions and
+    /// rc files, so a reviewer that never sees this is judging a different
+    /// command than the one that executes. Deliberately has no default: a caller
+    /// must state the environment rather than inherit a benign-looking one.
+    environment: command_environment.Environment,
 };
 
 pub const FileMutationAction = struct {
@@ -471,6 +478,30 @@ const SerializedEvidence = struct {
     }
 };
 
+/// Names the interpreter in the evidence, and marks the evidence incomplete when
+/// a shell-routed environment cannot say which shell. Incomplete action evidence
+/// already fails closed to `.invalid`, so an unnameable interpreter cannot be
+/// auto-approved.
+fn writeEnvironment(
+    writer: *std.Io.Writer,
+    alloc: std.mem.Allocator,
+    environment: command_environment.Environment,
+    action_complete: *bool,
+) !void {
+    try writer.print("environment: {s}\n", .{@tagName(std.meta.activeTag(environment))});
+    switch (environment) {
+        .legacy, .workspace_clean => {},
+        .clean, .user => |shell_path| {
+            if (shell_path.len == 0) {
+                action_complete.* = false;
+                try writer.writeAll("environment_shell: <unknown>\n");
+                return;
+            }
+            try writeBoundedField(writer, alloc, "environment_shell", shell_path, max_action_field_bytes, action_complete);
+        },
+    }
+}
+
 fn serializeEvidence(
     alloc: std.mem.Allocator,
     request: ReviewRequest,
@@ -502,6 +533,7 @@ fn serializeEvidence(
                 "background: {}\ntarget_os: {s}\n",
                 .{ command.background, @tagName(command.target_os) },
             );
+            try writeEnvironment(&out.writer, alloc, command.environment, &action_complete);
         },
         .file_mutation => |file| {
             try out.writer.writeAll("action: prepared_file_mutation\n");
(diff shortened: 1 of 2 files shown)


src/core/permissions/auto_classifier.zig
```diff
@@ -492,6 +502,30 @@ const SerializedEvidence = struct {
...
+    try writer.print("environment: {s}\n", .{@tagName(std.meta.activeTag(environment))});
+    switch (environment) {
+        .legacy, .workspace_clean => {},
+        .clean, .user => |shell_path| {
+            if (shell_path.len == 0) {
+                action_complete.* = false;
+                try writer.writeAll("environment_shell: <unknown>\n");
+                return;
+            }
+            try writeBoundedField(writer, alloc, "environment_shell", shell_path, max_action_field_bytes, action_complete);
```
DarlingtonDeveloper, line 501: **[P1] Include the actual shell invocation semantics**

`environment: user` and `/bin/bash` still omit the behavior that makes #237 exploitable. Fx actually executes `/bin/bash --login -O expand_aliases -c ...` or `/bin/zsh -l -i -c ...`; a reviewer given only `/bin/bash` can reasonably model ordinary `bash -c`, where aliases are disabled. The evidence nevertheless remains marked complete. Please serialize the exact invocation argv/startup flags—ideally derived from `capturedInvocation`—and test that the payload includes the login and alias-expansion semantics.

DarlingtonDeveloper (changes requested):

beardthelion:
Fixed in 24586cc. Agreed that the path alone lets a reviewer model an ordinary `bash -c`, where aliases are off.

The evidence now carries the startup words, derived from capturedInvocation rather than restated:

    environment_invocation: '/bin/bash' '--login' '-O' 'expand_aliases' '-c'
    environment_invocation: '/bin/zsh' '-l' '-i' '-c'

The zsh line makes your point better than the bash one: `-i` is what makes that shell expand aliases at all, and nothing about the path would tell a reviewer that. The command keeps its own field, so only the startup words appear here, and an invocation that cannot be resolved marks the evidence incomplete rather than describing a shell fx cannot build.

The evidence test asserts the expansion semantics are present; reducing it back to the path makes it fail.

One interaction worth flagging since it spans two PRs: #259 stops alias expansion for commands the auto-mode allowlist approves. Those never reach the reviewer, so this packet stays accurate either way, and the two changes do not overlap.

Suite is 8410 passing with the same pre-existing failures as the branch base.

beardthelion:
Correction on the SHA above: that commit is now 16ba5f1. I rebased onto main to pick up "Remove OS command sandboxing", which deleted SandboxWideningAction and the scope and backend fields this branch had touched. The widening half of the change went with it, since there is no longer a widening action to describe; the command evidence is unchanged.

8378 passing against an 8375 baseline on ebff2fe, same pre-existing failures.

PR #242, opened 2026-08-21
Edit composer prompts in an external editor

possibilities:
## Summary

- Add `Ctrl+T` to edit the current plain-text composer draft with `VISUAL`, falling back to `EDITOR`, including while a response continues in the background.
- Spawn the editor without a shell, limit and validate returned UTF-8 text, and retain a recovery file whenever a saved result cannot be used.
- Yield a clean cooked terminal to the child, protect in-flight worker syscalls from terminal interrupts, and restore raw mode, keyboard and paste modes, layout, and a full repaint on every return path.

## Scope and shortcut direction

`Ctrl+T` is provisional to avoid colliding with the existing `Ctrl+G` update/reload shortcut. `Ctrl+G` is a better fit for consistency with popular coding agents; a focused follow-up should move external editing to `Ctrl+G` and remove or relocate the current update binding.

This first version declines to open the editor while the draft contains pasted-block placeholders, image attachments, or skill tokens. That keeps structured entities intact without adding an offset-mapping subsystem. A successful external replacement resets composer undo history.

## Related work

PR #148 mentioned external-editor takeover as possible future work.

## Verification

- `zig fmt --check src/`
- `zig build`
- `zig build -Dwasm-surface=core -Doptimize=ReleaseSmall`
- `zig build -Dwasm-surface=term -Doptimize=ReleaseSmall`
- `zig build -Dnapi-surface=core -Doptimize=ReleaseSafe`
- clean-`ZDOTDIR` `zig build test --summary all`: 8,476 passed and 2 skipped, plus 1 benchmark test passed
- `tests/e2e/tui-composer-edit-contracts.test.ts`: 35 passed, including background worker progress while the editor is blocked, permission parking, terminal-mode restoration, and streaming guards
- direct-write audit and E2E shard-planner tests: 11 passed
- PGSO corpus validation passed; the new cases remain verification-only in the already classified composer contract file because editor handoff and recovery are important but uncommon behavior
- exercised the freshly built `./zig-out/bin/fx` in a real PTY while a held response completed behind a blocked editor; the answer and edited draft appeared after restore, fx stayed alive, and stderr remained empty

Required label: `type: feature`

Marked ready because Full CI passed for `0de69da` on our fork: https://github.com/possibilities/fx/actions/runs/32575054416

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ fx
 
 The current directory becomes the primary workspace. Enter a prompt, or run `/help` to browse interactive commands.
 
+Press `Ctrl+T` to edit the current plain-text prompt with the command in `VISUAL`, falling back to `EDITOR`. Returning from the editor replaces the prompt and resets its edit history. Drafts containing pasted blocks, images, or skills stay unchanged in fx.
+
 The status line hides the workspace path and Git branch by default. Enable the `Status line workspace` option in `/settings`, run `/statusline workspace`, or set it in `~/.fx/settings.json`:
 
 ```json
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ fx
 
 The current directory becomes the primary workspace. Enter a prompt, or run `/help` to browse interactive commands.
 
-Press `Ctrl+T` to edit the current prompt in the editor named by `VISUAL`, falling back to `EDITOR`; drafts containing pasted blocks, images, or skills are left unchanged.
+Press `Ctrl+T` to edit the current prompt with `VISUAL`, falling back to `EDITOR`, even while a response is streaming; drafts containing pasted blocks, images, or skills are left unchanged.
 
 The status line hides the workspace path and Git branch by default. Enable the `Status line workspace` option in `/settings`, run `/statusline workspace`, or set it in `~/.fx/settings.json`:
 
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ fx
 
 The current directory becomes the primary workspace. Enter a prompt, or run `/help` to browse interactive commands.
 
-Press `Ctrl+T` to edit the current plain-text prompt with the command in `VISUAL`, falling back to `EDITOR`. Returning from the editor replaces the prompt and resets its edit history. Drafts containing pasted blocks, images, or skills stay unchanged in fx.
+Press `Ctrl+T` to edit the current prompt in the editor named by `VISUAL`, falling back to `EDITOR`; drafts containing pasted blocks, images, or skills are left unchanged.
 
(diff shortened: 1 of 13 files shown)


possibilities:
Follow-up work is underway to allow `Ctrl+T` external editing while an agent turn continues in the background.

possibilities:
Additional changes are now pushed in c65cdc4. Ctrl+T can open the editor while a response continues in the background. This update also mirrors the full terminal handoff and restore lifecycle, uses restartable SIGINT/SIGQUIT handlers for in-flight worker syscalls, and adds deterministic coverage for concurrent worker progress, permission waits, terminal mode restoration, and modal or structured-draft guards. The adversarial review is clear with no remaining findings. The PR remains draft pending the required `type: feature` label, exact-commit Full CI on all four native runners, and the final ship gate.

possibilities:
Rebased onto `4a8f765` and force-pushed — this was 78 commits behind and `CONFLICTING`, so it was not reviewable as it stood.

The only conflict was in `tests/e2e/tui-composer-edit-contracts.test.ts` against d094c9f ("Retire legacy terminal presentation modes"), and it was mechanical: that commit dropped `maxxing_mode: "legacy"` from the harness settings while this branch had changed the same line to spread a per-test `settings` object, so the resolution is `JSON.stringify({ ...settings })`. The `❯` → `┃` prompt-glyph assertions took d094c9f's side wholesale; none of the cases added here asserted the old glyph.

Since d094c9f is exactly the kind of change that merges clean and still breaks a patch, I checked the handoff path against it rather than trusting the rebase: this branch works at the termios/layout level and reuses `footer_rows` (`src/main.zig:187`), the same constant upstream's own `suspendToJobControl` resume path uses, and it never referenced `input_appearance`, `maxxing_mode`, or `presentation_mode`. Nothing it calls was retired.

Post-rebase on this machine (macOS, aarch64):

- `zig fmt --check src/`, `scripts/check-public-surface.sh`
- `zig build`, `zig build -Doptimize=ReleaseSafe`
- `zig build -Dwasm-surface=core -Doptimize=ReleaseSmall`, `-Dwasm-surface=term`, `-Dnapi-surface=core -Doptimize=ReleaseSafe`
- `zig build test -Doptimize=ReleaseSafe` with a clean `ZDOTDIR`: green
- `tui-composer-edit-contracts.test.ts`: 35 pass / 0 fail
- `ci-shards.test.ts`: 7 pass; `tui-direct-write-audit.test.ts`: 4 pass

Still draft, and the open question from the description is unchanged: `Ctrl+T` is provisional, and `Ctrl+G` is the better home for external editing if the current update/reload binding can move.

PR #244, opened 2026-08-21
Preserve candidate transcript bounds during resume

possibilities:
## Summary

- Project a pending resume transcript against the candidate viewport bounds when they remain inside the terminal
- Preserve incremental resume publication when the candidate transcript extends one row below the stale base `content_bottom`
- Add a regression covering recovery through the previously rejected transition

## Root cause

A valid candidate layout could extend the transcript one row below the base layout's stale `content_bottom`. Projection validated against the stale bound and raised `InvalidTranscriptTransition`, even though the candidate remained within the terminal's actual rows.

## Testing

- `zig fmt --check src/`
- focused transcript recovery test
- resumed a 186x45 session that previously failed with `InvalidTranscriptTransition`, submitted draft input, and exited successfully with empty stderr

Required label: `type: bug`

Marked ready because Full CI passed for `d7efe6f` on our fork: https://github.com/possibilities/fx/actions/runs/32575059044

### DIFF
--- a/src/ui/transcript/runtime.zig
+++ b/src/ui/transcript/runtime.zig
@@ -7101,6 +7101,21 @@ pub const TranscriptRuntime = struct {
         ));
     }
 
+    fn layoutForTranscriptProjection(
+        layout: Layout,
+        target_area: render_engine.frame_layout.FrameRect,
+    ) !Layout {
+        if (target_area.bottom > layout.rows) {
+            return error.InvalidTranscriptTransition;
+        }
+        var projection_layout = layout;
+        projection_layout.content_bottom = @max(
+            projection_layout.content_bottom,
+            target_area.bottom,
+        );
+        return projection_layout;
+    }
+
     const TransitionTarget = struct {
         body_disposition: TranscriptBodyDisposition = .paint,
         selection: ViewportSelection,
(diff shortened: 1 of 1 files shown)


possibilities:
Rebased onto `4a8f765` (was 78 commits behind) and marked ready for review.

No conflicts, but `src/ui/transcript/runtime.zig` is a file d094c9f ("Retire legacy terminal presentation modes") also rewrote, so a clean rebase is not on its own evidence the patch still holds. What that commit removed from this file was the `maxxing_mode` field on `TranscriptRuntime` and the legacy-mode tests around it; this change touches neither. It hoists the `content_bottom` clamp into `layoutForTranscriptProjection` and adds the `target_area.bottom > layout.rows` guard that turns an out-of-bounds projection into `error.InvalidTranscriptTransition` instead of a bad paint. Pure geometry, no presentation coupling.

Post-rebase on this machine (macOS, aarch64):

- `zig fmt --check src/`, `scripts/check-public-surface.sh`
- `zig build`, `zig build -Doptimize=ReleaseSafe`
- `zig build test -Doptimize=ReleaseSafe` with a clean `ZDOTDIR`: green
- `tui-resume.test.ts`: 48 pass; `tui-resume-brutal.test.ts`: 1 pass; `tui-full-transcript-brutal.test.ts`: 2 pass

PR #245, opened 2026-08-21
Prevent Codex usage reservation exhaustion

possibilities:
## Summary

- Distinguish direct Codex completions that do not produce Gateway generation identities
- Settle usage reservations when generation observation fails, without recording duplicate incidents
- Add regressions for direct Codex `resp_*` identities and more than 64 consecutive completions

## Root cause

Direct Codex responses expose `resp_*` response IDs, but `GatewayCompletion.generation_id` accepts only Gateway `gen_*` identities. The rejected identity marked billing incomplete without releasing its active reservation, so the 65th successful response failed with `UsageCapacityExceeded`.

## Testing

- `zig fmt --check src/`
- focused session-usage and direct-Codex tests
- all Codex-focused tests
- exercised the built binary against a mock Codex endpoint for 66 responses without `UsageCapacityExceeded`

Required label: `type: bug`

### DIFF
--- a/src/core/agent/runtime/gateway_step.zig
+++ b/src/core/agent/runtime/gateway_step.zig
@@ -139,6 +139,7 @@ pub fn streamGatewayCompletion(
     const status = result.status;
     const finish_reason = result.completion.finish_reason;
     const completion_usage = result.completion.usage;
+    const gateway_generation_expected = result.completion.gateway_generation_expected;
     const delivery_ambiguous = result.completion.delivery_ambiguous;
     const generation_metadata_invalid = result.completion.generation_metadata_invalid;
     const provider_result_identity_failure = result.completion.provider_result_identity_failure;
@@ -148,6 +149,7 @@ pub fn streamGatewayCompletion(
         .completion = .{
             .content = content,
             .tool_calls = tool_calls,
+            .gateway_generation_expected = gateway_generation_expected,
             .generation_id = generation_id,
             .generation_metadata_invalid = generation_metadata_invalid,
             .delivery_ambiguous = delivery_ambiguous,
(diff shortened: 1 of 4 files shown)


reyanshgupta:
I independently reproduced this on fx `0.0.5` and opened #282

Before this patch, one successful saved Codex request produced:

```text
usage generation checkpointed incomplete reason=InvalidGenerationId
usage generation queued sequence=1 id=resp_[redacted]
```

and persisted `next_sequence: 2`, `settled_through_sequence: 0`.

I cherry-picked `81ee55f` onto current `main` (`df7e624`) and ran:

- `zig fmt --check src/`
- the three focused regression tests added by this PR
- the freshly built `./zig-out/bin/fx` against the real Codex subscription route in a saved session

The built binary returned `OK` with clean stderr and persisted:

```json
{
  "billing": "complete",
  "next_sequence": 2,
  "settled_through_sequence": 1,
  "api_duration_complete": true,
  "pending": [],
  "incidents": []
}
```

possibilities:
Superseded by #291 — closing this in favor of it.

#291 fixes the same reservation leak but declares `observes_gateway_usage` per provider route on `agentStream(.codex)`/`agentStream(.grok)`, rather than the per-completion `gateway_generation_expected` flag this PR added to `GatewayCompletion`. That is the better placement: the property belongs to the route, not to each response off it, and stating it there covers the Grok path this PR left untouched. #291 also carries both a `for (0..65)` unit test and a `CODEX_CAPACITY_65_OK` E2E, so the past-the-ceiling boundary this PR was built around is covered on both providers.

@reyanshgupta — thank you for the independent repro on `0.0.5` and for filing #282; I've pointed that trace at #291 so it lands with the fix that ships. #282 stays open until #291 merges.

Nothing here is worth keeping on top of #291.

PR #259, opened 2026-08-21
Run an allowlisted command as it was parsed

### LINKED ISSUE #258: Auto-approved commands are re-read by the operator's shell before they run
### Summary

In auto mode, a command approved by the safe-command allowlist is then re-read by the operator's login shell, so the command that runs can differ from the command that was approved. No reviewer is involved, and the transcript still shows the original string.

### The invariant

`knownReversibleAutoCommand` states its own contract:

> Returns true only for a small parsed set of ordinary development commands whose effects are expected and recoverable in auto mode. Unknown syntax, dynamic shell expansion, publication, deletion, and wrappers remain on the normal review path.

Alias expansion is shell expansion, applied to the same string after that parse. So the effect that runs is not the effect that was parsed, and the exclusion the comment describes is what the execution layer then performs.

### The path

`resolveOrdinaryPermissionOutcome` consults `knownReversibleAutoCommand` in its auto branch and returns `shellPermissionOutcome(.auto_mode)` before `automaticReviewOutcome` is reached, so the reviewer is never called. `reversibleGit` accepts `git status` (`src/core/shell_command/command_effect.zig:130`). Execution resolves `.user` as a login shell and, for bash, appends `-O expand_aliases` (`capturedInvocation` in `src/core/terminal/shell_resolver.zig`), which is what lets the profile redefine the command.

### Reproduction

With `alias git='touch ALIAS_EXECUTED'` in the profile of an isolated HOME, auto mode, one `git status` tool call:

```
alias present: classifier requests 0, tool status success, exit 0, ALIAS_EXECUTED created
no alias:      classifier requests 0, exit 128 (git itself, outside a repository)
```

Trace from the same run: `approval_source=auto_mode decision=once` and `sandbox explicit command environment=user shell=/bin/bash`.

`ask` mode blocks it: the command never runs. The sandbox setting made no difference here, though its backend looks inert on Linux and I have not tested macOS seatbelt.

### What this is and is not

The aliases are the operator's own configuration, not remote input, so this is not remote code execution and I would not call it a vulnerability. What it is: an approval decision made about a different command than the one that runs, in the mode whose purpose is to skip human review. The consequence worth weighing is durability. Anything that writes one line into a shell profile, including an approved edit fx makes itself, turns every allowlisted command into unreviewed execution from then on, and nothing in the transcript shows it.

Existing fixtures miss it because the isolated HOME they build has no shell profile, so there is no alias to shadow anything.

### Environment

Zig 0.16.0, Linux aarch64, reproduced on `195b73f`.

Related but distinct: #237 is about reviews that happen not naming the interpreter. This is about commands that never reach a reviewer.

beardthelion:
Fixes #258.

Auto-mode allowlist approvals now execute using the selected shell without allowing operator startup definitions to reinterpret the approved command. Bash uses `--noprofile --norc -p`; zsh uses `-f`.

Reviewed commands, configured rules, session grants, and interactive approvals retain the operator's normal shell definitions.

The contributor's original commit and attribution are preserved.

### Verification

- `zig build`
- focused Zig tests: 13 passed
- `auto-mode-reliability.test.ts`: 23 passed
- built-binary regression covers the captured login shell and confirms the automatic reviewer was not consulted

Full CI passed on exact commit `9abbb5d`: [run 32566855312](https://github.com/vercel-labs/fx/actions/runs/32566855312).

### DIFF
--- a/src/core/terminal/shell_resolver.zig
+++ b/src/core/terminal/shell_resolver.zig
@@ -197,20 +197,31 @@ pub fn capturedInvocationResolving(
                 .clean_start = true,
             } });
             removeInteractiveFlag(&invocation);
+            if (words == .as_parsed and shellKind(path).? == .bash) {
+                appendBashAsParsedProtection(&invocation);
+            }
             invocation.setCommand(command);
             return invocation;
         },
         .user => |path| {
+            if (words == .as_parsed) {
+                var invocation = try resolve(null, .{ .executable = .{
+                    .path = path,
+                    .clean_start = true,
+                } });
+                removeInteractiveFlag(&invocation);
+                if (shellKind(path).? == .bash) {
+                    appendBashAsParsedProtection(&invocation);
+                }
+                invocation.setCommand(command);
+                return invocation;
+            }
+
             var invocation = try resolve(path, .user_login);
             if (std.mem.eql(u8, std.fs.path.basename(path), "bash")) {
                 removeInteractiveFlag(&invocation);
-                // Alias expansion is what lets a login shell rewrite the command
-                // words. It stays on for a command a human or the reviewer looked
-                // at, and comes off for one approved by parsing the string.
-                if (words == .shell_defined) {
-                    invocation.append("-O");
-                    invocation.append("expand_aliases");
-                }
+                invocation.append("-O");
+                invocation.append("expand_aliases");
             }
             const effective_command = if (shellKind(path) == .zsh)
                 try std.mem.concat(alloc, u8, &.{ captured_zsh_user_prelude, command })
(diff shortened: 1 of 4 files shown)


src/core/terminal/shell_resolver.zig
```diff
@@ -183,8 +197,13 @@ pub fn capturedInvocation(environment_value: Environment, command: []const u8) R
             var invocation = try resolve(path, .user_login);
             if (std.mem.eql(u8, std.fs.path.basename(path), "bash")) {
```
DarlingtonDeveloper, line 221: **[P1] Enforce `as_parsed` for zsh too**

This only disables Bash alias expansion. zsh still runs with `-l -i -c`, allowing `.zshrc` aliases/functions to rewrite auto-allowlisted command words on the default macOS shell. Make `as_parsed` bypass zsh user definitions too, or execute the resolved program directly, and add a zsh alias regression.

DarlingtonDeveloper (changes requested):

DarlingtonDeveloper (approved):
The zsh path now bypasses user startup definitions, Bash also blocks BASH_ENV/imported functions via `-p`, reviewed commands retain user-shell behavior, and exact commit `9abbb5d` passed all four Full CI platform suites.

Issue #110, opened 2026-08-18
Respect the XDG base directory specification on Linux


mrakh:
It looks like the coding agent puts all its stuff under `~/.fx`, which unfortunately follows a bad tradition of coding agents polluting the home directory with their own dotfolders. Applications should instead respect [the XDG base directory specification](https://specifications.freedesktop.org/basedir/latest/) for storing application data:

* Configs under `${XDG_CONFIG_HOME}/fx`, or `${HOME}/.config/fx` if the `XDG_CONFIG_HOME` env is unset
* State (like session data and logs) under `${XDG_STATE_HOME}/fx`, or `${HOME}/.local/state/fx` if the `XDG_STATE_HOME` env is unset

I can create a PR for this change. It can be made non-breaking by checking first for the existence of `~/.fx` and using it if it's already there, and following the XDG convention otherwise.

giuseppeg:
@DarlingtonDeveloper landed here to ask for `FX_HOME `(mirroring `CODEX_HOME`) it is key (many cases i won't expand to keep this short).

Also would it be possible to change the layout of `sessions/` to group them by project folder like Claude Code does? In local environments this would allow to give read/write access only to the current cwd transcripts via things like @anthropic-ai/sandbox-runtime deny/allow fs policies.

PR #278, opened 2026-08-21
Respect the XDG base directories on Linux

arthjean:
Resolves the fx profile through the three XDG base directory roots on Linux, appending `fx` to each.

Refs #110

| Root | Variable | Default | Holds |
| --- | --- | --- | --- |
| config | `XDG_CONFIG_HOME` | `~/.config/fx` | `settings.json`, `mcp.json`, `AGENTS.md`, `backups/` |
| state | `XDG_STATE_HOME` | `~/.local/state/fx` | `sessions/`, `logs/`, credentials, `terminal-host/` |
| data | `XDG_DATA_HOME` | `~/.local/share/fx` | `skills/`, `memories.json` |

An unset, empty, or relative value is ignored in favor of the default. Missing base components are created 0700, and one that already exists keeps its mode.

Two rules override that resolution:

* A legacy `~/.fx` holding a recognized entry collapses all three roots back onto it. Nothing is moved, copied, or deleted to migrate.
* Only Linux splits. macOS and every other target keep `~/.fx` whatever the environment exports.

The terminal host moves only its socket to `XDG_RUNTIME_DIR` and keeps its identity in the state root, so a runtime directory cleared at logout never takes the host with it. `fx doctor` reports the active layout and the three resolved roots.

Full CI for the exact head commit passes on all four platforms. The macOS arm64 PGSO candidate failed two p95 tail-latency comparisons while the corresponding p50 changes remained below the 10% threshold. It also failed on a tui-subagent-manager.test.ts case that passed after the bounded retry in Full CI for the same commit.


arthjean:
Both red jobs look like known PGSO noise rather than regressions from this change.

**`Startup status`** compares the PGO candidate against the non-PGO control built from the same commit, so this PR's code sits on both sides and cancels out. It rejected at p95 +18.0% against a 10% tolerance, driven by 2 of 100 candidate samples at 24.47ms and 28.61ms against a 2.73ms median; no control sample exceeds 7.08ms. Trimming those two gives p95 -2.7%.

**`Verify behavior-16`** fails force close with `session_lost` while the trace reads `incomplete=false` on the descendants leg, which pins it to the shell-group leg of `terminalSignalCompleted` signaling an already-exited shell. `signalTrackedProcessWith` exempts `error.ProcessNotFound` from `incomplete`; that leg has no equivalent exemption. Same failure on `main` in run 32376866751 (Aug 20).

Across 9 completed runs on this branch PGSO produced 8 distinct failure signatures, spanning seed build, training, startup, heavy and behavior stages. Repo-wide it is 21 green to 37 red since Aug 16.

This PR touches none of `native_session.zig`, `process_tree.zig` or `terminal-host.test.ts`. PGSO runs here only because classifying the new `tests/e2e/profile-layout.test.ts` in `corpus.json` is mandatory.

Issue #282, opened 2026-08-22
Direct-subscription responses exhaust session usage reservations


reyanshgupta:
## Issue details

Saved sessions using the direct Codex subscription route leak one usage reservation per successful completion. After 64 completions, the next request fails locally with:

```text
System: request failed: UsageCapacityExceeded
```

This is not a ChatGPT subscription limit. The direct Codex response ID uses the `resp_*` namespace, while the session usage ledger accepts only Vercel AI Gateway `gen_*` generation IDs. Validation rejects the Codex ID, marks billing incomplete, and leaves the invocation reserved.

### Environment

- fx `0.0.5` (`df7e6245e199`)
- macOS arm64
- Provider: Codex subscription
- Model: `gpt-5.6-sol`
- `fx doctor`: authentication and startup checks pass

### Minimal reproduction

1. Sign in with `fx login codex`.
2. Run a saved request:

   ```sh
   FX_MODEL=gpt-5.6-sol fx ask --json --yolo "Reply with exactly OK."
   ```

3. Inspect the new session's `usage-v2.json`.

The request succeeds and returns `OK`, but its snapshot remains unsettled:

```json
{
  "billing": "incomplete",
  "next_sequence": 2,
  "settled_through_sequence": 0,
  "request_count": 0,
  "models": [],
  "pending": []
}
```

A reviewed trace contains:

```text
[session] usage generation checkpointed incomplete reason=InvalidGenerationId
[session] usage generation queued sequence=1 id=resp_[redacted]
```

Longer affected sessions reached `next_sequence: 65` and `next_sequence: 129` while `settled_through_sequence` remained `0`. Once 64 live reservations accumulate in one process, `GatewayObservation.begin` returns `UsageCapacityExceeded`. Restarting temporarily clears the in-memory reservations, but the leak resumes.

### Expected behavior

A successful direct-subscription completion should settle its usage reservation without requiring a Vercel AI Gateway generation ID. Invalid generation metadata must also never leave an active reservation behind.

### Existing fix

Draft PR #245 appears to address this exact failure by distinguishing direct Codex completions from Gateway generations and defensively settling rejected observations:

- https://github.com/vercel-labs/fx/pull/245

Related retry issue #230 covers repeated transport failures, but not this deterministic `resp_*` identity mismatch and 64-reservation exhaustion.

## Trace

The relevant reviewed and redacted trace lines are included above. No credentials, full response IDs, prompts containing private data, or private paths are included.

## Sensitive information

[...]

PR #291, opened 2026-08-22
Prevent subscription requests from exhausting Gateway usage

fazxes:
## Summary

- Keep Codex and Grok sessions usable beyond 64 sequential provider calls.
- Preserve Gateway usage observation and reconciliation behavior.
- Settle rejected Gateway generation metadata without publishing or billing that generation.

### DIFF
--- a/src/builtins/providers.zig
+++ b/src/builtins/providers.zig
@@ -1,3 +1,4 @@
+const std = @import("std");
 const model_provider = @import("../core/config/model_provider.zig");
 const model_catalog = @import("../core/gateway/model_catalog.zig");
 const stream_provider = @import("../core/agent/stream_provider.zig");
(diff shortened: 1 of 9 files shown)


DarlingtonDeveloper (approved):

possibilities:
Cross-linking #245, which fixes the same reservation leak on the Codex path only. I'm closing mine in favor of this one — the route-level `observes_gateway_usage` declaration on `agentStream(.codex)`/`agentStream(.grok)` is the better model than the per-completion `gateway_generation_expected` flag I put on `GatewayCompletion`, because it states the property where it's actually true rather than re-deriving it per response, and it covers Grok without a second patch.

Two things from #245's thread worth carrying over here:

- @reyanshgupta independently reproduced the Codex case on `0.0.5` and filed #282, with the `usage generation checkpointed incomplete reason=InvalidGenerationId` trace from a real session rather than a constructed one.
- My E2E asserted the 65th completion specifically because 64 is the reservation ceiling, so the first failure is silent until exactly one past it. Your `for (0..65)` unit test plus the `CODEX_CAPACITY_65_OK` E2E already cover that boundary on both providers, which is strictly more than mine did.

Nothing in #245 is worth salvaging on top of this.

PR #296, opened 2026-08-22
Complete cleanly ended Grok streams without terminal events

gbarros-dev:
The Grok CLI proxy can end a stream after it has already delivered tools or text without sending `response.completed`. fx treated that clean end as incomplete even when the delivered output was coherent.

## Summary

- Complete Grok SSE after clean EOF or `[DONE]` when every tool has parseable arguments, or when visible text arrived with no tools.
- Preserve `ReadFailed` as a transport failure so interrupted streams cannot silently commit truncated output.
- Keep empty, reasoning-only, and truncated tool streams as `XaiGrokStreamIncomplete`.
- Cover a valid delta followed by a partial later event and `ReadFailed`.

## Validation

- `zig fmt --check src/`
- `zig build test`
- `zig build`
- `./zig-out/bin/fx -v`
- `./zig-out/bin/fx --help`

The live Grok proxy close was not reproduced in this checkout. This PR handles proven clean stream endings; generic transport `ReadFailed` remains a failure.

GPT-5.6 Sol on Codex, with a clean Grok 4.6 review spike.

### DIFF
--- a/src/gateway/xai_grok.zig
+++ b/src/gateway/xai_grok.zig
@@ -569,11 +569,7 @@ fn consumeSse(
     var saw_content_delta = false;
     var event_count: usize = 0;
 
-    stream: while (true) {
-        const json_text = sse.next(alloc, reader) catch |err| switch (err) {
-            error.ReadFailed => break :stream,
-            else => return err,
-        } orelse break :stream;
+    while (try sse.next(alloc, reader)) |json_text| {
         defer sse.release();
         if (cancel_flag.load(.seq_cst)) return error.Cancelled;
         event_count = try checkedAccumulatedSize(event_count, 1, max_sse_events);
@@ -740,7 +736,7 @@ fn consumeSse(
     };
 }
 
-/// Grok's proxy sometimes closes after tool or text deltas without
+/// Grok's proxy sometimes ends cleanly after tool or text deltas without
 /// `response.completed`. Salvage only when every tool already has parseable
 /// arguments, or when the stream delivered visible text and no tools.
 fn streamHasUsableOutput(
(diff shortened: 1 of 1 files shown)


src/gateway/xai_grok.zig
```diff
@@ -569,7 +569,11 @@ fn consumeSse(
     var saw_content_delta = false;
     var event_count: usize = 0;
 
-    while (try sse.next(alloc, reader)) |json_text| {
+    stream: while (true) {
+        const json_text = sse.next(alloc, reader) catch |err| switch (err) {
+            error.ReadFailed => break :stream,
```
DarlingtonDeveloper: **[P1] Do not salvage transport read failures as completed output**

`ReadFailed` is collapsed into clean stream termination, after which earlier text or parseable tool arguments become `.stop`/`.tool_calls`. A connection dying midway through a later event silently commits truncated output. Preserve transport-failure state and limit salvage to proven clean EOF/`[DONE]`; add a valid-delta-plus-partial-event regression.

gbarros-dev: Addressed in 583b7a5. `ReadFailed` now propagates as a transport failure; only clean EOF or `[DONE]` reaches the usable-output salvage. I also added a regression with a valid text delta, a buffered partial later event, and `ReadFailed`. The PR description now states that the live proxy close remains unreproduced, so this change does not conflate that uncertainty with clean termination.

gbarros-dev:
Closing because the observed incident was a transport `ReadFailed`. After preserving that failure correctly, this PR only handles clean EOF or `[DONE]` without `response.completed`, which we did not reproduce. We should revisit this only with a captured transport trace that distinguishes explicit completion, clean EOF, and read failure.

PR #302, opened 2026-08-22
Wait between MCP legacy stream reconnects

### LINKED ISSUE #301: Legacy Streamable HTTP MCP server can drive fx into a no-delay reconnect loop
A legacy Streamable HTTP MCP server that closes a stream cleanly gets reconnected to with no delay at all, for as long as the session lives. Both the notification listener and the tool-call resume path are affected.

## Reproduction

Point fx at a legacy Streamable HTTP MCP server (any of 2025-11-25, 2025-06-18, 2025-03-26) that advertises `tools.listChanged`, and have its GET notification endpoint answer `200 text/event-stream` with a body that is already finished. Keep the session open.

Measured on `ea90c07` with a local fixture, fx binary built from that commit:

| case | requests | window |
|---|---|---|
| notification GET, stream closes immediately | 27482 | 5.0s |
| notification GET, server sends `retry: 1` then closes | 2164 | 3.0s |
| tool-call resume, server keeps closing a resumable stream | 16953 | 5.0s |
| notification GET, stream held open (control) | 1 | n/a |

The third case comes from a single `tools/call`, and it is bounded only by the operation deadline.

## Why the existing backoff does not catch it

`readNotificationStream` treats `EndOfStream` as a clean result and returns `retry_ms = 0` (`src/core/mcp/legacy_streamable_http.zig:608-617`). `notificationGetCore` guards its only wait, and with it the only cancellation and deadline check, behind `retry_ms > 0` (`:505-528`), so it reissues the GET immediately. Nothing errors, so the exponential backoff in `src/core/mcp/tool_subscription.zig:561-570` never runs, and the listener deadline is not a bound either: `listener_lifetime_ms` is `std.math.maxInt(u32)`, about 50 days (`tool_subscription.zig:18`).

`readSseWithResumption` (`legacy_streamable_http.zig:939-958`) has the same shape on the POST tool-call path.

The other two transports are not affected: both return an error on `EndOfStream` (`legacy_http_sse.zig:518`, `streamable_http.zig:941`), so the close reaches the outer backoff.

## Impact

A remote server, hostile or merely broken, decides fx's outbound request rate. It is not a hang, since cancellation still works and the session ends cleanly, but it burns CPU and hammers the server for the life of the session.

I have a fix and regression tests ready and will open a PR against this.

beardthelion:
Fixes #301.

A legacy Streamable HTTP MCP server that closes a stream cleanly was reconnected to with no delay, for the life of the session. A clean close is not an error, so the exponential backoff in `tool_subscription.zig` never saw it, and the listener deadline is `maxInt(u32)`, so nothing else bounded it either. The same shape existed on the tool-call resume path in the same file.

Both loops now wait one second between reconnects when the server sent no `retry` hint, and check the control deadline every iteration rather than only when the server asked for a delay. A hint the server did send is used as given. SEP-1699 makes the retry field a MUST, so the one second is a default and never a floor over an explicit value.

Measured with the new fixtures, before and after:

| case | before | after |
|---|---|---|
| listener, stream closes immediately | 14598 reconnects / 3s | <= 10 |
| tool-call resume, server keeps closing | 15838 / 5s | <= 10 |
| listener, stream held open (control) | 1 | 1 |

The other two transports already return an error on `EndOfStream` (`legacy_http_sse.zig:518`, `streamable_http.zig:941`), so their closes reach the outer backoff and they needed no change.

## Tests

Three fixture modes (`closed_listener`, `retry_hint_listener`, `resume_storm`) and four cases in `mcp-legacy-remote.test.ts`. Each assertion was checked against the unfixed binary first, so none of them is vacuously green:

- Reconnect spacing, not just a count in a window. A bare count passes under any delay, including one too small to matter; the spacing assertion fails at 0ms without the default.
- A lower bound alongside the upper one. Disabling the listener makes the test fail with `Received: 0`, which is what keeps the upper bound honest.
- The listener still delivers. From its second attempt the fixture sends `tools/list_changed` before closing, and the test asserts the tool list is refetched, so "stopped storming" cannot be confused with "stopped listening".
- The resume loop honors a short operation deadline between attempts, exiting in about two seconds with a cancellation sent upstream.
- An explicit `retry: 500`, the value the conformance suite uses, is honored rather than replaced. Observed spacing is 501ms across seven reconnects, and the assertion bounds it on both sides at the suite's own too-early and very-late thresholds.

The MCP conformance suite is 471 passed / 0 failed, with `sse-retry` on 2025-11-25 reporting `client-sse-retry-timing SUCCESS` and the GET landing 504ms after the stream close. The full `mcp-legacy-remote` suite is green at 41 tests. `mcp-http` and `mcp-stdio` match the pre-change baseline on the same commit, including their pre-existing failures.

## Trade-off worth naming

Against a server that sends no hint the wait is flat, so a well-behaved server that closes its stream cleanly waits up to a second before the listener reconnects. An escalating backoff would be gentler on a persistently broken server but adds state to a loop that currently has none, and one second is already three orders of magnitude below the behavior being fixed. A server that does send a hint sets its own pace, including a very short one, which is what the spec asks for.

The no-hint default is back to one second here. 500ms was picked while the value was a floor over the hint and had to clear the conformance window; now that the hint is honored the default only governs servers that sent nothing, so it is an independent choice. Happy to take 500ms instead if you prefer it.

### DIFF
--- a/src/core/mcp/legacy_streamable_http.zig
+++ b/src/core/mcp/legacy_streamable_http.zig
@@ -502,6 +502,12 @@ fn notificationGetCore(alloc: Allocator, options: OperationOptions) !OperationRe
     const sink = options.notifications orelse return error.MissingNotificationSink;
     var last_event_id: ?[]u8 = null;
     defer if (last_event_id) |value| alloc.free(value);
+    // The SSE retry field sets the reconnection time until the server changes
+    // it, so it outlives the stream that carried it. Holding it here is what
+    // keeps a hint sent once from being replaced by our own default on the
+    // next reconnect. Zero means no hint has arrived yet.
+    var hinted_ms: u32 = 0;
+    var reconnects: u32 = 0;
 
     while (true) {
         var next_options = options;
@@ -509,17 +515,25 @@ fn notificationGetCore(alloc: Allocator, options: OperationOptions) !OperationRe
         const resumed = try notificationGetOnceCore(alloc, next_options, sink);
         if (last_event_id) |value| alloc.free(value);
         last_event_id = resumed.last_event_id;
+        if (resumed.retry_ms > 0) hinted_ms = resumed.retry_ms;
         // A server-supplied retry is honored as given: SEP-1699 makes it a
         // MUST, and the MCP conformance suite fails a client that reconnects
-        // either earlier or more than twice as late as the hint. The default
-        // applies only when the server sent no hint at all, which is the case
-        // that storms: a stream that closes as soon as it opens reconnects
-        // with no wait, and no outer backoff sees it, because a clean close is
-        // not an error and the retry path in tool_subscription never runs.
-        const retry_ms = if (resumed.retry_ms > 0)
-            resumed.retry_ms
+        // either earlier or more than twice as late as the hint.
+        //
+        // Without a hint the first reconnect is immediate, because a stream
+        // that closed once after doing its work is the ordinary case and
+        // should not pay a penalty. From the second onward the default
+        // applies, which is what bounds the storm: a stream that closes as
+        // soon as it opens would otherwise reconnect with no wait forever,
+        // and no outer backoff sees it, because a clean close is not an error
+        // and the retry path in tool_subscription never runs.
+        const retry_ms = if (hinted_ms > 0)
+            hinted_ms
+        else if (reconnects == 0)
+            0
         else
             default_reconnect_delay_ms;
+        reconnects +|= 1;
         const now = std.Io.Clock.Timestamp.now(io_mod.getIo(), .awake);
         if (controlExpired(options.control)) {
             return error.McpRequestTimedOut;
@@ -944,17 +958,27 @@ fn readSseWithResumption(
     request_id: u64,
 ) ![]u8 {
     var stream = try readSseStream(alloc, initial_reader, options, request_id);
+    // Held across resumes for the same reason as the listener above: the retry
+    // field sets the reconnection time until the server changes it, so a hint
+    // sent on one stream still governs the next one.
+    var hinted_ms: u32 = 0;
+    var resumes: u32 = 0;
     while (true) switch (stream) {
         .final => |body| return body,
         .resumable => |resume_state| {
-            // Same default as the notification listener above, for the same
-            // reason: a server that keeps closing a resumable stream without a
-            // retry hint would otherwise be resumed with no wait at all. A
-            // hint the server did send is still honored as given.
-            const retry_ms = if (resume_state.retry_ms > 0)
-                resume_state.retry_ms
+            if (resume_state.retry_ms > 0) hinted_ms = resume_state.retry_ms;
+            // Same rule as the notification listener above: an explicit hint
+            // is honored as sent, the first resume is immediate because a
+            // stream that closed once after doing its work is ordinary, and
+            // the default from the second onward is what bounds a server that
+            // keeps closing a resumable stream without ever finishing.
+            const retry_ms = if (hinted_ms > 0)
+                hinted_ms
+            else if (resumes == 0)
+                0
             else
                 default_reconnect_delay_ms;
+            resumes +|= 1;
             {
                 const now = std.Io.Clock.Timestamp.now(io_mod.getIo(), .awake);
                 if (controlExpired(options.control)) {
--- a/src/core/mcp/legacy_streamable_http.zig
+++ b/src/core/mcp/legacy_streamable_http.zig
@@ -10,7 +10,7 @@ const streamable_http = @import("streamable_http.zig");
 const Allocator = std.mem.Allocator;
 const max_sse_events: usize = 1024;
 const close_timeout_ms: i64 = 250;
-const min_reconnect_delay_ms: u32 = 1000;
+const min_reconnect_delay_ms: u32 = 500;
 
 pub const Version = enum {
     v2025_11_25,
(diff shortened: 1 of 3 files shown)


src/core/mcp/legacy_streamable_http.zig
```diff
@@ -508,23 +509,30 @@ fn notificationGetCore(alloc: Allocator, options: OperationOptions) !OperationRe
...
-            try wake.wait(io_mod.getIo());
+        // Every reconnect waits at least min_reconnect_delay_ms, whether the
+        // server asked for a delay or not. Without the floor, a stream that
+        // closes as soon as it opens reconnects with no wait at all and the
+        // listener becomes a request storm that no outer backoff can see: a
+        // clean close is not an error, so the retry path in tool_subscription
+        // never runs. Clamping rather than defaulting matters because the
+        // delay is server-supplied, so `retry: 1` would otherwise buy the same
+        // storm the floor exists to stop.
+        const retry_ms = @max(resumed.retry_ms, min_reconnect_delay_ms);
```
DarlingtonDeveloper: **[P1] Do not override an explicit SSE retry interval**

This changes a server-supplied `retry: 500` into 1000 ms. Full CI confirms the regression: MCP conformance `ClientRespectsRetryField` fails because reconnection occurs at about 1002 ms instead of 500 ms. Preserve explicit retry values and apply a default/backoff only when the field is absent, or address the storm outside the protocol hint.

DarlingtonDeveloper (changes requested):

src/core/mcp/legacy_streamable_http.zig
beardthelion: > Preserve explicit retry values and apply a default/backoff only when the field is absent, or address the storm outside the protocol hint.

Done in 8076f9e, on top of your 4057030b. The wait is a default rather than a floor now: `retry_ms > 0` is used as the server sent it, and the one second applies only when there is no hint, which is the case #301 came from.

I reproduced your failure locally before touching anything, then confirmed the fix against the same runner: `sse-retry` on 2025-11-25 gave `Client reconnected very late (1003ms instead of 500ms)`, and now reports `client-sse-retry-timing SUCCESS` with the GET landing 504ms after the stream close. Full suite is 471 passed, 0 failed.

Worth flagging on your commit, since I rebased over it rather than replacing it. Lowering the floor to 500 clears the check but leaves the override in place below that. I ran the fixture with a 200ms hint against both builds:

| | observed gaps |
|---|---|
| 4057030b | 501, 501, 500, 501, 501, 500 |
| 8076f9e | 200, 201, 201, 200, 201 |

The suite passes at 500 because its `retryValue` is 500 and the floor was set to the same number. At `retry: 200` its very-late threshold is 400ms, so a 501ms reconnect would fail it. That is what pushed me to honor the field instead of picking a floor value.

Two things to sanity-check on your side. I put the no-hint default back to one second, since once the hint is honored that constant only governs servers that sent nothing; say the word if you want 500 there. And I inverted `retry_hint_listener`, which previously sent `retry: 1` and asserted it got clamped, so it now sends the suite's own 500 and bounds the observed spacing on both sides. Measured 501ms across seven reconnects. That puts the regression in `mcp-legacy-remote` rather than only in Full CI, which is why my default check rollup was green while this was broken.

The no-hint storm cases are unchanged and still fail without the default: 0ms listener spacing and 9519 resume requests. Each call site is covered separately, the listener by the e2e test and the resume path by conformance, and I checked that by clamping one site at a time.

beardthelion: You were still right about this one, and I was wrong to say it was done.

Honoring the hint fixed the stream that carried it and nothing after. `retry_ms`
is rebuilt per attempt in `readNotificationStream` and `readSseStream`, so the
value never survived into the next reconnect, and the retry field sets the
reconnection time until the server changes it rather than for one stream. A
server that sends `retry: 500` once and then omits it got 500ms on the first
reconnect and our 1000ms on every one after. I changed the fixture to send the
hint once and measured 1004ms, against the ~1002ms you originally reported. Same
substitution, one reconnect later.

The reason neither of us saw it is that `retry_hint_listener` repeats the field
on every response, so no fixture in the file could express a hint that stops
arriving. Dropping the stickiness now turns the new test red and leaves the old
one green, which is what the gap looked like.

The resume path had the same shape and no hint coverage at all: the only fixture
emitting a hint sat in the GET branch, so a floor or an override at that call
site passed all four tests. Overriding it now fails only the new test.

Fixed in ca4979a2 and a466e428, on top of your 4057030b:

- the interval is held across attempts at both call sites
- `retry_hint_once` covers a hint sent once and then omitted
- `retry_hint_resume` covers the resume path's hint branch
- the delay decision moved into one function both loops call, with a unit test,
  since two copies with one of them untested is how this got through

Two judgment calls worth your disagreement.

The first reconnect no longer waits when there is no hint. An ordinary stream
that closes after delivering its work was paying a full second: the three
per-version resume tests ran in 3.41s against 558ms without the delay, and the
2025-11-25 polling test 1139ms against 143ms. The default from the second
reconnect onward is what bounds the storm, and removing it still turns both
storm tests red. I did not gate on whether the attempt delivered events, which
was my first instinct, because the resume-storm server sends a progress
notification every time it closes and that gate would have reopened #301.

An explicit `retry: 0` falls back to the default rather than being honored
literally, since zero wait is the defect. Say the word if you would rather it be
honored as sent.

I also did not add a lower bound on small hints. `retry: 1` still reconnects at
that rate, unchanged from before this PR, but every bound I could write is an
override of an explicit interval, which is the thing you asked me not to do.
Happy to take it in a separate change if you want it.

Conformance is unchanged at 484 passed / 5 failed, with the same
`sep-2322-client-request-state` that also fails on an unmodified `main` here.
`mcp-legacy-remote` is 46/0 against 39/0 on the base. The branch is rebased onto
current `main`; CI on the new head has not finished yet.

PR #312, opened 2026-08-22
Expose structured access to terminal sessions

DarlingtonDeveloper:
## Summary

- expose a structured session API from `createFxTerminal()` backed by the same live Fx instance
- publish versioned state, conversation, tool, notice, attention, and turn lifecycle events
- preserve one worker FIFO while enforcing terminal-owned interaction and API-owned cancellation boundaries
- document the JavaScript API and cover shared history, state, rendering, and cancellation end to end

## Testing

- `zig fmt src/`
- `zig build test`
- `zig build -Dwasm-surface=term`
- `npm run test:term`
- `zig build`
- fresh `./zig-out/bin/fx` interactive smoke test

## Tracking

FXC-261


possibilities:
Very cool. As an aspirational follow-up, a native side channel attached to each fx process would be really useful for a project I’m working on—perhaps reusing this attached-session design over an inherited pipe or Unix socket to bring structured control to native, headful instances.

FWIW, I'm making something like a fx-specific multiplexer/ADE. Long way to go but basics live here https://github.com/possibilities/fmx

PR #332, opened 2026-08-22
Release the usage invocation slot when a generation identity is rejected

### LINKED ISSUE #333: Bug: Codex sessions fail with UsageCapacityExceeded after 64 requests
### Issue details

**Version:** fx 0.0.5, macOS arm64, provider `codex` (ChatGPT subscription, `gpt-5.6-sol`).

Partway through a Codex session every prompt starts failing with:

```
System: request failed: UsageCapacityExceeded
```

It never recovers inside that session. Quitting fx and resuming works — for another 64 requests, then it happens again.

**Steps to reproduce**

1. `fx login codex`, `fx provider codex`
2. Run 65 requests in one session (any prompts).
3. From request 65 on, every prompt fails with `request failed: UsageCapacityExceeded`.

**Cause**

The error is not a provider signal — it is the session usage ledger's own active-invocation cap (`max_active_invocations = 64` in `src/core/session/session_usage.zig`). `GatewayObservation.begin` reserves a slot per Gateway call, and `finishObservedInvocation` validates the generation identity *before* releasing it:

```zig
try validateGenerationId(id);   // returns here for Codex ids
...
if (!self.finishInvocationUnlocked(sequence, duration_ms, outcome)) return;
```

`validateGenerationId` accepts only Gateway ids (`gen_` + 26 ULID characters), while Codex returns OpenAI response ids (`resp_…`). So every Codex request strands its reservation, and request 65 fails in `reserveInvocation`. Restarting clears it because `restore` resets `active_sequence_count` to 0.

A session that had hit this four times shows the fingerprint exactly (`next_sequence: 257` = 4 × 64 reservations, nothing ever accounted):

```json
{"billing":"incomplete","api_duration_complete":false,"api_duration_ms":0,
 "next_sequence":257,"settled_through_sequence":0,"request_count":0,"models":[]}
```

Two side effects of the same path: API duration is never accounted on Codex sessions, and one `incident: incomplete` record is appended to `~/.fx/usage.jsonl` per request.

Fix proposed in #332.

### Trace

One `fx ask` request on Codex:

```
[session] usage generation checkpointed incomplete reason=InvalidGenerationId
[session] usage generation queued sequence=1 id=resp_02dd7a…
```

mercho40:
## Problem

On a Codex (ChatGPT subscription) session, every prompt starts failing with:

```
System: request failed: UsageCapacityExceeded
```

Once it starts, it never recovers within that session. Restarting fx and resuming works — for another 64 requests.

`UsageCapacityExceeded` here is not a provider signal, it is the session usage ledger's own active-invocation cap (`max_active_invocations = 64`).

## Root cause

`GatewayObservation.begin` reserves an invocation slot per Gateway call and the slot is released by `finishInvocationUnlocked`. On the observed path, identity validation ran first:

```zig
pub fn finishObservedInvocation(...) !void {
    try validateGenerationId(id);   // returns here
    ...
    if (!self.finishInvocationUnlocked(sequence, duration_ms, outcome)) return;
```

`validateGenerationId` accepts only Gateway ids (`gen_` + 26 ULID characters). Codex returns OpenAI response ids (`resp_…`), so the early return stranded the reservation on every single request:

```
[session] usage generation checkpointed incomplete reason=InvalidGenerationId
[session] usage generation queued sequence=1 id=resp_02dd7a185a8e269d016a89d27cbe7487d2ac741377642d1acf
```

The 65th `reserveInvocation` then returns `error.UsageCapacityExceeded`, which propagates out of the prompt as a system error. Restarting clears it because `restore` resets `active_sequence_count` to 0.

A session that hit this four times shows the fingerprint exactly — `next_sequence: 257` (4 × 64 reservations), with no invocation ever accounted:

```json
{"billing":"incomplete","api_duration_complete":false,"api_duration_ms":0,
 "next_sequence":257,"settled_through_sequence":0,"request_count":0,"models":[]}
```

## Fix

* `finishObservedInvocation` finishes the invocation before an identity failure surfaces, so a rejected identity can never strand active capacity. The outcome is recorded as `possibly_billed_without_identity`, matching what the ledger already concluded through the error path.
* `GatewayObservation.complete` routes an id that is not a Gateway generation id through the same path as a missing id, instead of pushing an unusable identity into the observed path. Ambiguous transport evidence keeps `.ambiguous_delivery` as the recorded reason.

Billing semantics are unchanged: these sessions were already recorded as incomplete, and the ledger still files one `incident: incomplete` per subscription request exactly as before. Whether a subscription request should instead count as `.unbilled` is a separate decision, raised on #333.

## Verification

`zig build`, then two new focused tests in `session_usage.zig`, each running `max_active_invocations + 1` iterations to prove capacity no longer drains:

* `rejected generation identity releases the invocation slot` — also asserts the API duration of every rejected invocation is accounted
* `unsupported provider generation identity keeps invocation capacity` — also covers an ambiguous delivery with an unsupported id, and that every sequence settles

Exercised with `./zig-out/bin/fx ask` on a Codex subscription. Trace before/after:

```
before: usage generation checkpointed incomplete reason=InvalidGenerationId
after:  usage billing incomplete sequence=1 reason=generation_identity_unsupported
```

Persisted snapshot for that request now accounts the invocation, where it previously recorded nothing:

```json
{"billing":"incomplete","api_duration_complete":true,"api_duration_ms":1689,
 "next_sequence":2,"settled_through_sequence":1}
```

Full local suite: 8424 passed / 5 failed on this branch versus 8422 passed / 5 failed on `main` — the same five pre-existing failures, which are shell-profile tests that pick up local `zsh` startup noise.

Required label: `type: bug` (I cannot apply labels on this repository).

Fixes #333.

### DIFF
--- a/src/core/session/session_usage.zig
+++ b/src/core/session/session_usage.zig
@@ -146,6 +146,27 @@ pub const GatewayObservation = struct {
             );
             return;
         }
+        if (!types.validGatewayGenerationId(id)) {
+            // Subscription providers return their own response identifiers.
+            // The ledger only reconciles Gateway generation ids, so finish the
+            // invocation without identity tracking rather than pushing an
+            // unusable identity through the observed path. Ambiguous transport
+            // evidence stays the recorded reason.
+            try ledger.finishInvocationDurably(
+                self.sequence,
+                self.elapsedMs(),
+                if (delivery == .ambiguous_delivery)
+                    delivery
+                else
+                    .possibly_billed_without_identity,
+            );
+            debug_trace.logf(
+                "session",
+                "usage billing incomplete sequence={d} reason=generation_identity_unsupported",
+                .{self.sequence},
+            );
+            return;
+        }
         try ledger.finishObservedInvocationDurably(
             alloc,
             self.sequence,
@@ -561,9 +582,21 @@ pub const Usage = struct {
         origin: []const u8,
         team: ?[]const u8,
     ) !void {
-        try validateGenerationId(id);
-        try validateOrigin(origin);
-        if (team) |value| try validateTeam(value);
+        validateObservedIdentity(id, origin, team) catch |err| {
+            // A rejected identity must not strand the reservation: the
+            // invocation is finished before the failure surfaces so active
+            // capacity cannot drain across a session.
+            self.mutex.lockUncancelable(io_mod.getIo());
+            defer self.mutex.unlock(io_mod.getIo());
+            _ = self.finishInvocationUnlocked(
+                sequence,
+                duration_ms,
+                .possibly_billed_without_identity,
+            );
+            self.billing = .incomplete;
+            self.dirty = true;
+            return err;
+        };
 
         self.mutex.lockUncancelable(io_mod.getIo());
         defer self.mutex.unlock(io_mod.getIo());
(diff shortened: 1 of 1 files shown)


mercho40:
Closing — superseded by #291, which merged with the route-level `observes_gateway_usage` declaration and covers Grok as well. Nothing here is worth keeping on top of it. Sorry for the noise; I filed before searching.

PR #335, opened 2026-08-22
Stage transcript projections against the rows the frame granted

### LINKED ISSUE #334: Bug: fx --resume aborts with InvalidTranscriptTransition on a healthy session
### Issue details

**Version:** fx 0.0.5, macOS arm64, provider `codex`.

One saved session cannot be resumed: fx exits immediately with

```
$ fx --resume <id>
fx: InvalidTranscriptTransition
```

The session data itself is fine — `fx session <id>` prints the whole 8-turn transcript, and `fx session recover <id>` correctly refuses ("the session has a valid commit boundary"). What fails is the first frame after the resume projection, and the error escapes to the top level and kills the process instead of degrading to a plain repaint.

**What I narrowed down**

* Deterministic: fails on every launch of that session.
* Not terminal-size dependent in the usual sense — it fails at 80x24, 100x30, 120x40, 160x50, 200x100. It only succeeds when the terminal is tall enough for the whole transcript to fit without scrolling (~950+ rows for this session, restored offset 0–5).
* Not a general resume regression: other saved sessions resume fine at non-zero restored offsets (68, 90, 193).
* Isolated to a single turn. Rebuilding the checkpoint with `history[0:7]` resumes cleanly at `target_visual_offset=798`; `history[7:8]` alone (one user turn, 64 tool steps, ~2.1 MB of state, several file-diff presentations with 1461-line before/after content) reproduces the failure at `target_visual_offset=81`.
* Independent of the `resume-view.bin` cache — reproduces with the cache deleted (`outcome=stale` / cold).

The failing check is inside `resolveTransitionTarget` (`src/ui/transcript/runtime.zig`), reached from `resolveTranscriptTransitionTargetForFrame`; `attempt_restore` then reports the error rather than falling back.

Two things seem worth separating:

1. Whatever makes this turn's projection produce a target the transition rejects.
2. A resume that cannot paint its restored viewport should fall back (repaint from the bottom, or drop the restored offset) rather than aborting the process — a rendering invariant currently makes an otherwise healthy session unopenable.

### Trace

```
[session] event=resume_view_cache outcome=stale
[session] event=session_replay source=checkpoint checkpoint_seq=5 tail_bytes=0 elapsed_us=84327
[session] event=resume_projection turns=8 project_us=101865 finalize_us=157988 install_us=794
[render]  viewport_prepared cursor=37,64 last_visible_row=37 replaceable_row=1 reserved_rows=0
[scroll]  transcript_target_resolution_failed err=InvalidTranscriptTransition source_compatible=true geometry_rebase=false recovery_rebase=false normal_buffer_recovery_pending=false target_visual_offset=973
[frame_schedule] attempt_restore outcome=error err=InvalidTranscriptTransition reasons=first_frame,transcript,external_damage reason_count=3 invalidations=1 pending_reason_count=3
```

Restored offsets by terminal height for the same session (120 columns): 24 rows → 1037 (fail), 40 → 973 (fail), 100 → 887 (fail), 900 → 113 (fail), 950 → 5 (ok), 1000 → 0 (ok).

mercho40:
## Problem

Some saved sessions cannot be reopened. `fx --resume <id>` exits immediately:

```
$ fx --resume <id>
fx: InvalidTranscriptTransition
```

The session is intact — `fx session <id>` prints the whole transcript and `fx session recover` correctly refuses because the commit boundary is valid. What fails is the first frame after the resume projection, and the error escapes the render attempt and ends the process, so the session is simply unopenable.

## Root cause

Transition staging measures the projection against a `types.Layout`, and the painter guards it:

```zig
if (prepared.selection.split_active or
    prepared.use_selected or
    area.isEmpty() or
    area.top == 0 or
    area.bottom > layout.content_bottom)
{
    return error.InvalidTranscriptTransition;
}
```

`area` is the transcript area **this frame** solved, while `self.layout.content_bottom` is the app layout's content bottom. They disagree whenever the frame reserves fewer footer rows than the app layout does — on the failing resume, the frame granted the transcript rows 1..37 while `content_bottom` was still 36:

```
[scroll] transition_resolve state=invalid target_offset=973 source_offset=0 total_rows=1010 area=1..37
[scroll] painter_reject site=stage_guard split=false selected=false empty=false top=1 bottom=37 content_bottom=36 rows=40 offset=64
[scroll] transcript_target_resolution_failed err=InvalidTranscriptTransition ... target_visual_offset=973
[frame_schedule] attempt_restore outcome=error err=InvalidTranscriptTransition reasons=first_frame,transcript,external_damage
```

Two of the staging sites in `resolveTransitionTarget` already handled this, rebasing the layout onto the frame's area before staging:

```zig
var projection_layout = self.layout;
projection_layout.content_bottom = target_layout.transcript_area.bottom;
```

The other five — the stable semantic-progress staging, the stable history projection, both recovering paths, and the invalid path — passed `self.layout` unchanged, so a frame whose transcript area reached past the app content bottom aborted the transition instead of staging into it.

## Fix

Resolve the projection layout once, from the rows the frame actually granted the transcript, and use it for every staging call in the transition. The two sites that already rebased keep exactly their previous behavior.

For the other five sites this changes two cases:

* **Frame area below the app content bottom** (the crash): the guard now passes and the projection stages into the frame's rows.
* **Frame area above the app content bottom** (a frame that reserves more rows than the app layout — an activity row, a banner): `content_bottom` now drops to the frame's bottom, so a terminated last line that fills the area exactly places the cursor on the area's last row instead of one row below it, outside the rows the frame granted. This matches what the two existing rebase sites already did; it is a deliberate alignment, not an accident of the fix.

### Why this layer

`self.layout` is the app's fixed-footer `types.Layout`; `frame_layout.solve` is the per-frame source of truth, and the two can legitimately disagree on where the transcript ends. The transition already acknowledged that by rebasing at two sites, so extending that to every staging call is the smallest change consistent with the existing design. Keeping `self.layout.content_bottom` in sync with the committed frame layout would remove the need for the rebase entirely; that is a broader change and out of scope here.

## Verification

`zig build`, plus a focused test — `frame transcript rows past the layout content bottom still stage` — that resolves a transition whose transcript area ends one row below `layout.content_bottom`. It fails with `error.InvalidTranscriptTransition` when `projectionLayout` returns the layout unrebased and passes with the fix.

Exercised with `./zig-out/bin/fx --resume` on the session that produced the report, driven through a pty at 80x24, 100x30, 120x40, 160x50 and 200x100: every size resumed, painted the transcript tail and status line correctly, and logged no transition rejection. Three other saved sessions resume unchanged.

Full local suite: 8425 passed / 5 failed, the same five pre-existing failures as `main` (shell-profile tests that pick up local `zsh` startup noise).

Fixes #334.

Required label: `type: bug` (I cannot apply labels on this repository).

### DIFF
--- a/src/ui/transcript/runtime.zig
+++ b/src/ui/transcript/runtime.zig
@@ -7286,6 +7286,17 @@ pub const TranscriptRuntime = struct {
         });
     }
 
+    fn projectionLayout(
+        self: *const TranscriptRuntime,
+        target_layout: render_engine.frame_layout.CommittedLayoutSnapshot,
+    ) Layout {
+        var layout = self.layout;
+        if (!target_layout.transcript_area.isEmpty()) {
+            layout.content_bottom = target_layout.transcript_area.bottom;
+        }
+        return layout;
+    }
+
     fn paintedStableTargetWouldRewind(
         self: *const TranscriptRuntime,
         target: TransitionTarget,
@@ -7365,6 +7376,12 @@ pub const TranscriptRuntime = struct {
         activity_overlay_active: bool,
     ) !TransitionTarget {
         var target = TransitionTarget.init(prepared, scroll_facts);
+        // Staging measures against the rows this frame granted the transcript,
+        // which need not match the app layout's content bottom: a frame that
+        // reserves fewer footer rows hands the transcript an area reaching
+        // past it, and measuring against the stale bottom rejects the
+        // transition outright.
+        const projection_layout = self.projectionLayout(target_layout);
         switch (self.transcript_commit_state) {
             .stable => |anchor| {
                 target.history_visual_offset = @min(
@@ -7410,7 +7427,7 @@ pub const TranscriptRuntime = struct {
                 )) {
                     try target.stagePreparedProjection(
                         alloc,
-                        self.layout,
+                        projection_layout,
                         prepared,
                         target_layout.transcript_area,
                         scroll_facts.source_visual_offset + accepted_semantic_progress_rows,
@@ -7422,7 +7439,7 @@ pub const TranscriptRuntime = struct {
                 {
                     try target.stagePreparedHistoryProjection(
                         alloc,
-                        self.layout,
+                        projection_layout,
                         prepared,
                         target_layout.transcript_area,
                     );
@@ -7449,8 +7466,6 @@ pub const TranscriptRuntime = struct {
                     // release: slide the window with an in-place repaint at
                     // the target offset. The rows passed over stay
                     // unreleased and settle later through the replay.
-                    var projection_layout = self.layout;
-                    projection_layout.content_bottom = target_layout.transcript_area.bottom;
                     try target.stagePreparedProjection(
                         alloc,
                         projection_layout,
@@ -7470,8 +7485,6 @@ pub const TranscriptRuntime = struct {
                     );
                 }
                 if (self.stableHistoryFloor(target, anchor, scroll_facts)) |history_floor| {
-                    var projection_layout = self.layout;
-                    projection_layout.content_bottom = target_layout.transcript_area.bottom;
                     try target.stagePreparedProjection(
                         alloc,
                         projection_layout,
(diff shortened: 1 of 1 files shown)


mercho40:
Closing in favour of #244, which fixes the same `content_bottom` mismatch, is already rebased onto the same base, takes the conservative `@max` form, and carries a regression through the real resume path. One data point for #244 from this side: the session behind #334 fails at every terminal size below ~950 rows and resumes cleanly once the projection layout is rebased, so it is a second independent reproduction of that change.

PR #338, opened 2026-08-22
Streamline terminal argument decoding

fazxes:
## Summary

- Reduce allocation count and latency when decoding complex terminal tool arguments.
- Preserve terminal action validation, null placeholders, stringified composite fields, and failure responses.

Stacked on #294.


fazxes:
Consolidated into #294.

PR #347, opened 2026-08-22
Keep the OAuth loopback callback listener alive through stray connections

Sachin1801:
## Summary

- Keep Codex and Grok OAuth loopback listeners alive through idle preconnects and unrelated browser requests.
- Keep callback listeners alive when browsers abort speculative connections with a TCP reset.
- Keep callback reads cancellable so a silent connection cannot block sign-in completion.
- Report authorization denials only after the callback state matches the active attempt.
- Keep Grok callbacks on an ephemeral loopback port.
- Allow the xAI accounts page to deliver Grok authorization codes through Chrome local-network callback checks.
- Accept pasted Grok authorization codes in interactive and CLI sign-in when the browser cannot return to fx.
- Show a styled browser completion or failure page after OAuth callbacks.
- Preserve safety caution labels when sessions resume.


DarlingtonDeveloper:
Thanks for the detailed investigation. Two questions before merging:

1. `awaitBrowserCallback` still performs a blocking read with a 30-second socket timeout, while the login poll’s request deadline is 15 seconds and cancellation joins the worker. The bare-preconnect test closes its socket immediately. Could we add a regression where an idle preconnect remains open while the real callback arrives, and ensure both the callback and cancellation remain responsive?

2. A stale denial with the wrong state is ignored, but a stale successful callback returns `*OAuthStateMismatch`, which terminates the current attempt. With stable callback ports, completing an old browser tab could therefore kill a retry. Should mismatched-state callbacks be answered with 404 and ignored while the listener continues waiting?

Sachin1801:
Both were real. Fixed in the two commits above.

## 1. The blocking read

You were right, and it was worse than a slow timeout.

`cancel_flag` was read at exactly two points, both inside `browserCallbackReady` around the 100 ms `poll(2)`. `accept`, the request read and the response write had no cancel check at all. So a preconnect that stayed open owned the worker thread for the full 30 s `SO_RCVTIMEO`, and since `SignInRuntime.cancel()` does an unbounded `thread.join()` and is called from the picker path in `auth_runtime.zig`, that stall landed on the UI thread. Not a slow cancel, a freeze.

The deadline interaction is worse than "30 s vs 15 s", too. `pollTokenStep` computes the 15 s deadline *before* calling `poll_device_token`, and for the browser flows that callee is the callback wait. Time spent waiting on a silent socket is subtracted from the token-exchange budget, so a stray connection did not just delay the login, it made the exchange that followed fail with `error.Timeout`.

The read is now poll-driven. Each accepted connection gets a 250 ms silence budget, consumed in 100 ms slices with `cancel_flag` checked on both sides of every slice; a connection that has not spoken by then is closed and the loop moves to the next one in the accept queue. The poll sits behind `reader.interface.bufferedLen() == 0`, so it costs one `poll` per socket read rather than one per byte, and a request already sitting in the reader's buffer is never waited on. `SO_RCVTIMEO` stays as an unreachable backstop.

One thing worth calling out, because it would have been a silent regression: once the read can return `error.Cancelled`, the bare `catch { ...; continue; }` that used to wrap it would have demoted cancellation into "stray connection, keep looping". That catch is now an explicit switch, which is also what makes `*OAuthCallbackTooLarge` reachable instead of swallowed.

Both regressions you asked for are in, per provider:

- **`...outruns an idle preconnect held open`** — a preconnect is opened and *held*, the genuine redirect is delivered on a second connection, and the test asserts the code comes back in under 1 s.
- **`...cancels while an idle preconnect is open`** — same held preconnect, `cancel_flag` flipped from another thread, asserts `error.Cancelled` in under 1 s.

The existing `CallbackProbe` could not express this — its empty request is `continue` under a `defer stream.close`, so the server got `EndOfStream` instantly and never reached the timeout. The new `HeldPreconnectProbe` holds the socket open on an atomic release flag, and only hands the listener control once the redirect bytes are already in the kernel buffer, so the elapsed-time assertions measure the listener rather than the probe.

I checked that both actually catch the bug rather than passing either way, by reverting each half of the fix separately. They turn out to guard different halves, which is worth knowing if either is ever touched:

| reverted | `outruns...` | `cancels...` | wall clock |
|---|---|---|---|
| silence budget -> 30 s | **FAIL** (elapsed) | pass, 435 ms | 47.8 s |
| 100 ms slicing -> one long wait | pass | **FAIL** (elapsed) | 47.5 s |
| neither (as committed) | pass | pass | ~5 s |

So the budget is what keeps the *callback* responsive and the slicing is what keeps *cancellation* responsive — the two halves of your question, and one test each.

Worth adding: when I tried to revert the readiness gate outright, it did not compile. `error.Cancelled` is only in `readBrowserCallbackTarget`'s inferred error set because the gate puts it there, so the `error.Cancelled => return err` prong loses its member and the compiler rejects it. The cancellation path is now structurally load-bearing rather than something a later edit can quietly drop.

Residual trade-off, stated plainly: a browser that meant to reuse its preconnect gets it closed and pays one extra loopback handshake. That is the ordinary idle keep-alive close every HTTP server does, every response here already carries `Connection: close`, and the reconnect is served by the same accept loop. Avoiding even that costs a multiplexed poll over the listener plus a carried set of partial connections — roughly 150 lines of state machine — which did not seem worth it for one handshake.

## 2. Mismatched state

Agreed, and done exactly as you suggested: 404 and keep waiting.

Worth naming *why* you are right, since it is a regression this PR created rather than a pre-existing wart. With ephemeral ports a stale tab redirected to a port nothing was bound to, so a mismatched success was mostly unreachable. Making the ports stable is what put an old tab and a live retry on the same listener — so the stable-port change is precisely what turned `*OAuthStateMismatch` from theoretical into "completing an old tab kills your retry".

`parseBrowserCallbackTarget` still returns the distinct error so it stays diagnosable and the parser-level test keeps its meaning; only `awaitBrowserCallback` changed, and a mismatch now takes the same path a mismatched denial already took, with a `debug_trace` line. Success and denial are symmetric again: a callback that cannot be proven to belong to this attempt is a stray, never fatal.

This is also the safer posture rather than a relaxation — the client still refuses to exchange a code it did not request, it just no longer lets anything that can reach loopback end a live login by hitting the port with the wrong state. New socket-level test per provider: `...ignores a stale success and waits for this attempt`.

## Two adjacent fixes in the same pass

Both are the same class of bug as the one this PR is about, so I did not want to leave them:

- `browserCallbackReady` mapped `POLLERR`/`POLLHUP`/`POLLNVAL` on the **listening** socket to `Invalid*OAuthCallback`. That call site is `if (!try browserCallbackReady(...))`, outside the catch that makes that error non-fatal — so a broken listener propagated as a fatal error wearing the wrong name. Now `*OAuthCallbackListenerFailed`.
- `readBrowserCallbackTarget` inferred "too large" from `request_len == request_bytes.len`, so a well-formed request whose terminator landed exactly on the last byte was misreported. It now tracks whether the terminator was actually found.

Also aligned `grok_oauth.zig`'s `bindBrowserCallback` with `chatgpt_oauth.zig`'s `bindBrowserCallback(e2e: bool)`. Grok had no such parameter, so its socket tests bound the real port 8976 and collided with any live `fx login grok` on the same machine; adding two more socket tests per provider would have made that worse.

## Verification

`zig fmt --check src/` and `./scripts/check-public-surface.sh` clean. `zig build test`: **8444/8451 passed, 2 skipped, 5 failed** — the same 5 pre-existing environmental failures as before this change (`tool_runtime`/`command_runner` shell tests picking up `(eval):1: can't change option: zle` from a local `~/.zshrc` under non-interactive zsh), none in `src/core/auth/`.

`grok_oauth.zig` and `chatgpt_oauth.zig` are still exact mirrors — `diff` of `browserRequestReadable` is empty, and of `readBrowserCallbackTarget` shows only the `Grok`/`ChatGpt` error-name substitutions.

Sachin1801:
> Thanks for the detailed investigation. Two questions before merging:
> 
> 1. `awaitBrowserCallback` still performs a blocking read with a 30-second socket timeout, while the login poll’s request deadline is 15 seconds and cancellation joins the worker. The bare-preconnect test closes its socket immediately. Could we add a regression where an idle preconnect remains open while the real callback arrives, and ensure both the callback and cancellation remain responsive?
> 2. A stale denial with the wrong state is ignored, but a stale successful callback returns `*OAuthStateMismatch`, which terminates the current attempt. With stable callback ports, completing an old browser tab could therefore kill a retry. Should mismatched-state callbacks be answered with 404 and ignored while the listener continues waiting?

I have made the fix for these as well, thank you for pointing out

PR #355, opened 2026-08-22
Add per-model AI Gateway provider routing

tyrailleverett:
Add per-model AI Gateway provider routing

### Summary

Adds a `providerRouting` key to `~/.fx/settings.json` that routes AI Gateway chat requests to specific upstream providers per model, using the Gateway's provider filtering and ordering (`order` / `only` provider slugs):

```json
{
  "providerRouting": {
    "default": { "order": ["wafer"] },
    "anthropic/claude-sonnet-5": { "only": ["baseten"], "order": ["baseten", "bedrock"] }
  }
}
```

- An entry keyed by model ID applies to that model; the `default` entry applies to every other model. Exact matches win over `default`.
- Routing options merge into the existing `providerOptions.gateway` request object alongside Fast mode's `speed`.
- Workspace overrides layer over profile-global entries per model.
- Codex and Grok subscription routes are unaffected; routing applies only to Gateway requests.

### Implementation

- New `src/core/config/provider_routing.zig`: owned per-model map with bounded parsing (entry/slug limits), lookup precedence, layer merge, and unit tests.
- Settings plumbing through `config_runtime.zig`, startup state, `App`, and both turn-config paths (interactive and `fx ask`).
- Orchestrator resolves routing against the active model at request build time.
- `gateway_json.writeProviderOptions` serializes `order` / `only` into the merged `gateway` object.
- README documents the new key.

### Testing

- Unit tests at every layer: parser validation and precedence, settings round-trip and diagnostics, request-body shapes (order-only, only-only, merged with Fast mode), and orchestrator flow tests proving routing reaches request bodies through the real pipeline.
- Verified end to end with a live AI Gateway request: trace confirmed `"only":["wafer"]` on a deepseek model request, accepted by the Gateway.
- `zig build test` passes; `zig fmt` clean. ReleaseSafe binary builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,19 @@ The status line hides the workspace path and Git branch by default. Enable the `
 }
 ```
 
+Gateway requests can be routed to specific upstream providers per model with `providerRouting` in `~/.fx/settings.json`. Provider slugs follow the AI Gateway's provider filtering and ordering. An entry keyed by model ID applies to that model; the `default` entry applies to every other model:
+
+```json
+{
+  "providerRouting": {
+    "default": { "order": ["wafer"] },
+    "anthropic/claude-sonnet-5": { "only": ["baseten"], "order": ["baseten", "bedrock"] }
+  }
+}
+```
+
+`order` lists the providers to attempt in sequence and `only` restricts routing to the listed providers. Workspace overrides in `~/.fx/settings.json` layer over profile-global entries per model.
+
 List saved sessions with `fx sessions`. Resume the latest session for the current workspace, or select an exact session ID, through the same command group:
 
 ```bash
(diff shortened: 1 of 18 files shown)


tyrailleverett:
Label required: `type: feature` (I don't have permission to apply labels on this repository).

khalido:
deepseek/deepseek-v4.1-flash is great but vercel has nine providers and their speed and answer quality vary a lot from request to request. I want DeepSeek first and the gateway's normal order as the fallback:

`{ "providerRouting": { "deepseek/deepseek-v4.1-flash": { "order": ["deepseek"] } } }`

In fact I am finding with both openrouter and vercel ai gateway with the cheaper good models that the provider order is super important.

Won't bore ppl with my own tests, but glm 5.3 flash and ds 4.1 flash can go from awesome and cheap to slow and crap depending on the provider. 

It's a bit of a wild west at the moment. I want these gateways to start doing some kind of quality control or checks. In the meantime, this kind of model config support is very important!

PR #356, opened 2026-08-22
Reduce MCP catalog memory

fazxes:
## Summary

- Reduce retained memory and allocation work as MCP tool catalogs grow.
- Keep MCP search results, schema selection, permissions, and protocol responses unchanged.


fazxes:
Consolidated into #294.

PR #364, opened 2026-08-23
Wipe request-side credentials instead of leaving them in freed memory

beardthelion:
The device-code, refresh and revoke flows each build a form body that holds the credential itself, and the JS host encodes a Bearer token into its header JSON. Each of those buffers went back to the allocator without being overwritten, while the response in the same function was carefully wiped with `secret.zeroAndFree`. The token stayed readable in freed heap memory after the call.

Overwriting the final buffer is not enough on its own, which is what makes this more than a one-line change. Every one of these paths grows a buffer while holding the secret. The form writer starts near 132 bytes and reallocates as the body grows, which a JWT-sized refresh token easily triggers, and `allocPrint` sizes the Bearer header generously before shrinking to the exact length. Each growth step abandons an allocation that keeps its own plaintext copy. A scan of released buffers found five per form call, not one.

So each path now reserves the encoded worst case up front, or sizes the allocation exactly, and only then overwrites the single remaining buffer at scope exit. The overwrite goes through a new `secret.zero` helper beside the existing `zeroAndFree`, so every secret overwrite still greps as one convention.

`requestDeviceAuthorization` keeps its plain writer, because its form carries only the client id and the scope.

## Verification

The oauth test drives all three form calls through an allocator that inspects each buffer as it is released. Two details make it load-bearing rather than decorative: its credential is long enough to force the growth path, and it searches for a short needle at the head of the value rather than the whole credential, since a buffer abandoned mid-growth holds only a truncated prefix. An earlier version of the test missed both and passed against the unfixed code.

Measured per call site, with a 2 KiB credential, before and after:

| site | before | after |
| --- | --- | --- |
| `refreshToken` | 5 | 0 |
| `revokeToken` | 5 | 0 |
| `pollDeviceTokenBounded` | 5 | 0 |
| `executeBearerGet` (JS host) | 2 | 0 |

Removing the reservations while keeping the overwrite puts the test at `expected 0, found 9`, so it covers the growth path specifically and not just the final buffer.

Local `zig build test` and `zig build test -Doptimize=ReleaseSafe` both leave the failing-test set identical to the pre-change baseline on this machine. ReleaseSafe matters here because dead-store elimination is the thing `secureZero` exists to survive, and a Debug-only result would not have shown that.

## Not covered

The JS host path cannot run without a host, so its reservation is guarded by a bound test rather than an end-to-end one. The end-to-end measurement above came from a standalone harness with the host imports stubbed.

`headersJsonCapacity` accepts both call shapes, a comptime tuple and a runtime slice, and the test exercises both. Only the tuple shape is reachable in a native build today, so without that coverage the slice branch would have compiled purely by not being analysed.

The response side is a separate problem and is not addressed here. Tokens parsed out of a JSON response stay in the parser's own released allocations, which I have filed separately. Those counts are identical before and after this branch.

I cannot apply labels on this repo. The intended one is `type: security`.

### DIFF
--- a/src/core/auth/chatgpt_oauth.zig
+++ b/src/core/auth/chatgpt_oauth.zig
@@ -451,8 +451,9 @@ fn refreshSession(
     session: *chatgpt_session.Session,
 ) !void {
     try mutation.requireWritable();
-    var body: std.Io.Writer.Allocating = .init(alloc);
+    var body: std.Io.Writer.Allocating = try .initCapacity(alloc, refreshBodyCapacity(session.refresh_token));
     defer body.deinit();
+    defer secret.zero(body.written());
     try body.writer.writeAll("{\"client_id\":");
     try std.json.Stringify.value(client_id, .{}, &body.writer);
     try body.writer.writeAll(",\"grant_type\":\"refresh_token\",\"refresh_token\":");
@@ -632,8 +633,15 @@ fn exchangeAuthorizationCodeForRedirectWithBounds(
     deadline: ?std.Io.Clock.Timestamp,
 ) !TokenSet {
     var form: FormBody = .{};
-    var body: std.Io.Writer.Allocating = .init(alloc);
+    var body: std.Io.Writer.Allocating = try .initCapacity(alloc, formCapacity(&.{
+        "grant_type",    "authorization_code",
+        "client_id",     client_id,
+        "code",          authorization_code,
+        "code_verifier", code_verifier,
+        "redirect_uri",  redirect_uri,
+    }));
     defer body.deinit();
+    defer secret.zero(body.written());
     try form.append(&body.writer, "grant_type", "authorization_code");
     try form.append(&body.writer, "client_id", client_id);
     try form.append(&body.writer, "code", authorization_code);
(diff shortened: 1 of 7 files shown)


beardthelion:
Extended this to the other two providers, which have the same defect and were not covered.

Grok and ChatGPT open their request writers with no reserved capacity, so the refresh token, the authorization code and the PKCE verifier are copied into every buffer the writer outgrows and each one goes back to the allocator intact. Five sites across `grok_oauth.zig` and `chatgpt_oauth.zig`, reserved and wiped the way the Vercel path already was. The ChatGPT refresh body is JSON rather than a form, so its bound comes from the six character `\uXXXX` worst case instead of percent-encoding.

Three `Authorization` headers were also still built with `allocPrint`, which sizes a buffer, discovers the real length and releases the first one with the token still in it: `grok_oauth.zig`, `login_flow.zig` and `mcp_auth.zig`. That was fixed by hand here in `js_host_auth.zig`, so the four sites now share one helper in `secret.zig`. The scan allocator moved there too, since three modules need it.

Measured with that allocator: four released blocks still held the verifier on the Grok code exchange and four on the ChatGPT one, zero after. Reverting any single reservation on its own brings its own count back, so none of the new tests is vacuous.

Two limits worth stating. The probe proves the reservation half only; in a safety build `Allocator.free` overwrites the final block before a wrapping allocator sees it, so the wipe at scope exit is not observable there. And the ChatGPT refresh body could not be attributed by scanning, because the auth file read and the JSON parse copy the same token in the same spelling, so that one is covered by measuring the reservation against the encoded length directly. The residual the scan does see on that path, two blocks per load, is the auth file parse, which is the same family as #365 on a different side and is untouched here.

Required label is `type: security`.

#437 touches `refreshSession` in both provider files. The only overlap is that both branches append tests to the end of the file, no logic collides, and with both test blocks kept the provider modules are green.

beardthelion:
Rebased onto current `main` (`8e3eabbc` -> `269545f1`). This branch was 315 commits behind, so its previous green CI ran against a base that no longer resembles `main`. It re-ran on the fresh base after the push.

The rebase was clean and the three commits are unchanged: `git range-diff` reports all three identical. Build and `zig fmt --check src/` pass, the Zig suite matches a same-commit `main` baseline with no new failures, and the eight credential-wiping tests this branch adds all pass, including the two allocator-probe ones that were the most likely to be disturbed by the jump.

This also still needs a `type: security` label and I can't apply one.

PR #385, opened 2026-08-24
gateway: reuse a process-wide HTTP client for streaming requests

ChloeVPin:
## What

`streamGatewayCompletionCoreWithOptions` constructed and destroyed a fresh `std.http.Client` per retry attempt, with `keep_alive = false` on the request. Every model request therefore paid full DNS + TCP + TLS connection setup — even mid-session, on every single turn.

This PR lazily initializes one process-wide client (mutex-guarded init; heap-allocated so it outlives any request arena) and keeps server-side connections alive so std.http.Client's connection pool can reuse them across retry attempts and chat turns.

## Why it matters

Measured connection-setup cost to the gateway host from a residential connection:

```
TCP+TLS handshake: 240 ms  (per model request, previously unavoidable)
DNS lookup:        ~190 ms first use
```

In an interactive session — where fx is a long-lived process handling many turns — this overhead was paid on **every turn**. With pooling, turn 2+ reuses the warm connection and pays zero setup. Retries within a turn also stop re-paying handshake costs.

## Notes

- The shared client intentionally uses `std.heap.c_allocator`, never a request arena, since it outlives individual requests.
- Init is race-safe via an atomic spin lock (std.Thread.Mutex is unavailable in Zig 0.16); contention is limited to first-use.
- Non-streaming metadata calls are untouched to keep the diff minimal; they can adopt the same client in a follow-up if maintainers want.
- `zig build -Doptimize=ReleaseFast` green; full test step passes; validated end-to-end with a live streaming request against the gateway.

Happy to adjust scope (e.g., extend pooling to the non-streaming helpers) if reviewers prefer.

### DIFF
--- a/src/gateway/client.zig
+++ b/src/gateway/client.zig
@@ -1154,6 +1154,37 @@ fn streamGatewayCompletionCore(
     );
 }
 
+/// Process-wide HTTP client for gateway streaming. Reusing one client lets
+/// std.http.Client's connection pool keep TCP/TLS connections warm across
+/// retry attempts and chat turns instead of paying DNS + TCP + TLS setup
+/// (~1-3 round trips, often 100-300 ms) on every model request.
+var shared_gateway_client: ?*std.http.Client = null;
+const shared_client_c = struct {
+    pub var mu: std.atomic.Value(bool) = .init(false);
+
+    fn tryLock() bool {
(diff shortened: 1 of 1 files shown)


ChloeVPin:
Withdrawing this PR — closing at the author's request.

PR #391, opened 2026-08-24
Fall back to an existing bash or zsh before /bin/bash

y0usaf:
When the account shell is not bash or zsh, login-shell resolution fell back to hardcoded `/bin/bash`. That path does not exist on NixOS, so captured commands failed with `FileNotFound` and terminal sessions wedged waiting on a shell that never started.

`fallbackLoginShell` now returns the platform default only when it exists on disk, otherwise probes `/usr/bin/bash` and `/run/current-system/sw/bin/bash` before giving up on the legacy default. Behavior is unchanged wherever `/bin/bash` or `/bin/zsh` exists; on hosts without them, resolution finally yields a real shell path.


y0usaf:
Required PR label per CONTRIBUTING.md: `type: bug` (I don't have permission to apply labels). Happy to flip to ready for review once Full CI completes on `8a2c7fc`.

Issue #276, opened 2026-08-21
Reasoning effort is silently ignored for gateway models


alphastorm:
## Summary

The `effort` setting has no wire effect for gateway models. On the default gateway route (`/v3/ai/language-model`), fx sends only `headers`, `prompt`, `toolChoice`, and `tools` — no reasoning or thinking configuration, no `temperature`, and no `maxOutputTokens` — regardless of the configured effort. Only the OpenAI Codex provider maps effort into its requests, so for every gateway model the documented setting is silently ignored and reasoning-capable models run at whatever the provider default happens to be.

Verified against the v0.0.5 release commit (`df7e624`).

## Observed behavior

Captured against a loopback fake gateway (no credentials, `FX_GATEWAY_BASE_URL`/`FX_GATEWAY_CHAT_URL`/`FX_E2E_GATEWAY_MODELS_URL` overrides), model `zai/glm-5.2`, `~/.fx/settings.json` containing `"effort": "high"`:

- request body top-level keys: `headers`, `prompt`, `toolChoice`, `tools` — nothing else
- identical body shape with effort unset, so the setting produces no observable difference on this route

## Source

- `src/gateway/openai_codex.zig:87` maps `provider_options.reasoning` into `"reasoning":{"effort":…}` for the Codex provider
- the gateway request build in `src/gateway/client.zig` has no equivalent: no reasoning, sampling, or max-output mapping for `/v3/ai/language-model`

## Impact

- `effort` (settings key, a profile-owned documented option) is a silent no-op for the default provider
- thinking-capable gateway models (for example GLM 5.x) cannot have reasoning enabled, disabled, or budgeted from fx; behavior is pinned to upstream defaults with no user control
- users comparing effort levels on gateway models measure noise

For clarity: this did not cause #233 — the decisive rerun there passed 3/3 with no fx changes once the executor worked (see #233 for receipts). Filing separately because it is an independent contract gap.

## Suggested direction

1. Map effort into the gateway request via the AI SDK protocol's provider options for families that support it, mirroring the Codex mapping.
2. Where a model or family cannot express effort, surface that (documentation and/or a one-time notice) instead of silently accepting the setting.

Reproduction is deterministic with the env overrides above; happy to share the capture harness.

## Probe evidence (gateway route, exact fx headers)

[...]

PR #400, opened 2026-08-24
Expose Fast mode over ACP

JonathanRiche:
## Summary

- expose a model-gated Fast mode select in session/new and session/load configOptions
- handle normal and fast through session/set_config_option and persist the preference
- reset Fast mode when switching to a model or provider model without Fast support
- cover serialization, durable rollback, persistence, prompt-wire behavior, validation, and model switching

## Why

ACP clients need the same Fast control that fx already exposes interactively. The local ACP schema has no boolean config-option variant, so this uses a select with normal and fast values and omits it for models whose catalog entry does not advertise Fast support.

This advances #276 without changing provider or gateway request construction.

## Testing

- zig test -ODebug --dep build_options -Mroot=src/acp/sessions.zig -Mbuild_options=.zig-cache/c/a9e9d0d980db0aac99bbb236ec7c4e53/options.zig --test-filter "Fast config option"
- focused server durable-rollback test through a temporary root importing src/acp/server.zig
- zig build -Dtarget=x86_64-linux-gnu
- bun test acp.test.ts -t "Fast config follows"
- zig fmt --check src/
- git diff --check

Built-binary proof used ./zig-out/bin/fx acp with initialize, session/new, Fast=fast, and session/prompt. It exited 0 with empty stderr; FX_TRACE recorded fast_mode=true and fast=selected on the provider-options event.

### Authenticated runtime proof

On the exact PR commit, a freshly built `./zig-out/bin/fx acp` used the authenticated Codex subscription with `gpt-5.6-sol`. `session/new` exposed `provider`, `model`, `fast`, and `mode`; selecting `fast` followed by a real prompt returned the requested marker with `stopReason=end_turn`. The process exited 0 with empty stderr.

### Grok runtime proof

An authenticated `grok-4.6` run confirmed the fail-closed catalog gate: because that model does not advertise Fast support, the ACP response omitted the `fast` option. A real prompt still completed with `stopReason=end_turn`, exit 0, and empty stderr.

### Vercel AI Gateway runtime proof

An authenticated Gateway run used `openai/gpt-5.2`, exposed the Fast option, selected `fast`, and completed a real prompt with `stopReason=end_turn`, exit 0, and empty stderr.

### Current-main verification

Rebased onto upstream `1b87677`. The exact rebased commit passed the 13-test focused Zig filter, the focused ACP E2E, `zig fmt --check src/`, `git diff --check`, and a fresh `x86_64-linux-gnu` build. A real authenticated Codex prompt through `./zig-out/bin/fx acp` selected Fast mode, returned `stopReason=end_turn`, exited 0, and produced empty stderr.

### DIFF
--- a/src/acp/server.zig
+++ b/src/acp/server.zig
@@ -1551,6 +1551,10 @@ fn handleSetConfigOption(state: *ServerState, alloc: Allocator, msg: *jsonrpc.Me
                 .code = ErrorCode.invalid_params,
                 .message = "Invalid session model",
             });
+        const next_fast_mode = session.fast_mode and sessions.modelSupportsFastMode(
+            value,
+            state.capability_resolver.catalogEntries(),
+        );
         if (comptime !host_target.is_wasm) {
             if (session.provider != .gateway) {
                 var model_available = false;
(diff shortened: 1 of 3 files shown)


JonathanRiche:
Maintainer action requested: please apply the single required label `type: feature`. GitHub rejects label changes from this contributor account, so the PR must remain draft under CONTRIBUTING.md until the label is present and exact-commit Full CI plus the ship gate pass.

PR #406, opened 2026-08-24
Steer active agent turns

TY37zhang:
## Summary

- Steer an active turn with Ctrl+Enter.
- Preserve Enter for ordinary queued follow-ups.
- Keep late steering as the next queued turn.

## Problem

- Active prompts could only wait as separate turns.
- Steering must not replace existing queue behavior.
- Completion races must not lose submitted input.

## Solution

- Consume steering at safe model boundaries.
- Preserve steering across queue review and session history.
- Show pending steering distinctly from queued prompts.


mjlbach:
Hey! I implemented this here if you'd like to review:

https://github.com/vercel-labs/fx/pull/368

DarlingtonDeveloper (changes requested):
This is the better-scoped steering implementation and it has an excellent CI/binary-size profile. Two correctness issues need fixing before merge:

1. `finishProcessing()` appends late steering after the ordinary queue, reversing submission order. A steer submitted before a queued follow-up executes after it. Please preserve one admission order, ideally using a unified queue with durable steer metadata.
2. `queuePreview()` includes steering, but `clearQueuedPrompts()` only clears ordinary prompts. After “Cancel all,” steering remains and may still be delivered.

Please add regressions for both cases. The unified ordering work in #368 is useful reference, without that PR’s unrelated scope.

DarlingtonDeveloper (approved):

PR #407, opened 2026-08-24
Keep agent turns alive after missing tool authority

TY37zhang:
## TL;DR
Keep agent turns running when an allowed tool lacks execution authority. The affected tool now fails closed while the model can recover.

## Issue

- Missing tool authority aborts the entire agent turn
- Serial and parallel paths share the failure mode

## Fix

- Convert missing authority into a failed tool result
- Preserve continuation after serial and parallel failures
- Add regression coverage for both execution paths


TY37zhang:
Closing because the reported v0.0.5 failure is already fixed on main by the automatic-review hardening changes. This defensive patch does not address an additional production issue.

PR #409, opened 2026-08-24
Stop terminal commands at their timeout

fazxes:
## Summary

- Stop foreground terminal process groups and every descendant fx can track when their requested timeout expires.
- Report the cleanup scope and mark fully detached macOS descendant cleanup as best effort.
- Keep cleanup working when commands clear their environment or run through Bash scripts.
- Prevent trailing shell statements and output after a timeout.
- Preserve the existing graceful cancellation behavior.
- Label terminal timeouts separately from command failures.
- Report exit codes, signals, wait ceilings, condition matches, output-capture failures, and structured terminal errors.
- Preserve terminal outcome labels after session resume.


DarlingtonDeveloper:
macOS still leaks detached descendants when a command uses env -i, which removes the tracking marker. In a 50-process double-fork/setsid repro, 26 remained alive after timeout.

fazxes:
Fixed in bf4c512 and 2f88a65. The env -i, setsid, and double-fork repro now passes 20/20 with no surviving descendants or delayed effects.

PR #421, opened 2026-08-25
Show the terminal command under its output block

bdsoliv:
## Summary

When the harness displays a terminal execution, the terminal output block now shows the command that was run, as a dimmed prompt row underneath the output (after the exit code / folded-hint row):

```
│ line-1
│ line-2
│ exit code 7
│ $ printf lines; exit 7
```

The command is extracted from the tool detail's `arguments_json`, terminal-safe encoded, and capped at 120 bytes, matching the status-line projection. It renders for live output and compact transcript projections (both `buildCommandOutputOverridesInterruptible` and `syncCommandOutputBlockEntries`).

## Implementation

- `command_output_runtime.zig`: `renderCompactCommandOutputWithProcessPresentation` accepts an optional command and renders it via `appendCompactTerminalRows`; new `capturedCommandForBlock` mirrors `processPresentationForBlock`'s detail matching and extracts the `command` argument.
- `source_preparation.zig` and `syncCommandOutputBlockEntries` pass the captured command through.
- Test coverage in the streamed-command runtime test asserts the command row renders after the exit code row.

## Verification

- `zig build` clean, `zig fmt --check` clean.
- Focused runtime test updated and passing logic verified against the render path; the full unit suite cannot complete locally (pre-existing first-test disk-I/O stall on the 460 MB test binary), so Full CI is required on the exact commit.

### DIFF
--- a/src/ui/transcript/command_output_runtime.zig
+++ b/src/ui/transcript/command_output_runtime.zig
@@ -17,7 +17,9 @@ const std = @import("std");
 const debug_trace = @import("../../core/shared/debug_trace.zig");
 const display_width = @import("../../core/shared/display_width.zig");
 const io_mod = @import("../../core/shared/io.zig");
+const text_utils = @import("../../core/shared/text_utils.zig");
 const types = @import("../../core/shared/types.zig");
+const tool_args = @import("../../core/tooling/tool_args.zig");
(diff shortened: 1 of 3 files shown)


bdsoliv:
Closing per request; this was opened without approval.

PR #451, opened 2026-08-26
Fix remote MCP discovery and response framing

fazxes:
## Summary

- Fall back to supported legacy MCP versions when discovery errors use nonnumeric response IDs.
- Keep successful MCP responses strict about matching request IDs.
- Read fixed-length modern MCP JSON and SSE responses without process aborts.
- Close each modern MCP request after one response, including early final SSE events.

### DIFF
--- a/src/core/mcp/streamable_http.zig
+++ b/src/core/mcp/streamable_http.zig
@@ -270,6 +270,11 @@ const PreparedRequest = struct {
     }
 };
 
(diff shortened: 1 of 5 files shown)


fazxes:
Superseded by #429. The discovery fallback and fixed-length response work now lives on the combined project-scoped MCP branch at 0305484e7.

PR #477, opened 2026-08-27
Say when the team list could not be loaded during sign-in

### LINKED ISSUE #476: Failed team lookup during login is reported as a successful sign-in
### What happens

If the team lookup fails during `fx login`, the sign-in reports success and quietly gives you a session scoped to your personal account.

There is no way to tell it happened. Against a fake issuer returning 500 for `/v2/teams`:

- exit code `0`
- stdout byte-identical to a healthy login, including `Signed in to Vercel.`
- stderr empty
- `auth.json` written with no `team_slug` and no `team_id`

A healthy login and a broken one differ only by two absent fields in a file most people never open.

### Why it matters

`Credential.gatewayTeam()` returns null without those fields, so no `teamId` is attached to gateway requests. The comment on `fetchCreditsWithFetch` in `src/builtins/gateway.zig` says `/v1/credits` "rejects it outright unless the request names one", and team-scoped private models stop being visible. The session persists that way, so it stays broken until the next login.

### Cause

Both sign-in paths in `src/core/auth/login_flow.zig` do:

```zig
var teams = fetchTeams(alloc, token.access_token, issuer_url) catch std.ArrayList(Team).empty;
```

once in `completeSignIn` (the in-app sign-in) and once in `runLogin` (the CLI). Every failure collapses into an empty list, which is exactly what an account with genuinely no teams produces, so nothing downstream can tell the two apart. `selectTeam` returns null for an empty list, no picker appears, and the success message prints unconditionally.

A related case reaches the same end without any error at all: `parseTeams` skips entries whose `id` or `slug` is not a string, so a `200` whose entries drift from the expected shape also yields an empty list, silently.

### Reproduce

Point `FX_E2E_OAUTH_ISSUER_URL` at a local issuer that serves the OAuth device flow normally but returns `500` for `/v2/teams`, then run `fx login`. Compare stdout and `~/.fx/auth.json` against the same run where `/v2/teams` returns a team. A `200` with body `{"teams":[{"id":123,"slug":"acme"}]}` reproduces the silent variant.

### Note

The sibling path treats this failure as significant: `loadTeamSelection`, used by `fx teams`, calls the same `fetchTeams` with `try` and fails loudly. Only the sign-in paths swallow it.

I have a fix and will open a PR shortly.

beardthelion:
Fixes #476.

Needs `type: bug`. I cannot apply labels, so this stays in draft until a maintainer adds it. Full CI is green on this commit across all four native runners.

A failed team lookup during sign-in was indistinguishable from an account that has no teams, so `fx login` reported success and quietly left you in personal scope. Both sign-in paths did:

```zig
var teams = fetchTeams(alloc, token.access_token, issuer_url) catch std.ArrayList(Team).empty;
```

Every failure became an empty list, which is exactly what a zero-team account produces. No picker, no message, `auth.json` written without `team_slug` or `team_id`, and stdout byte-identical to a healthy login.

The sign-in itself is fine, so this keeps it. The token pair is valid and the team fetch is auxiliary; failing the login over it would be worse, and re-running the device flow is expensive. What was missing was any signal, so that is what this adds.

### What changed

Both call sites now track whether the list actually arrived. On a caught error the CLI prints one line after the existing gateway-access caveat, and the in-app flow emits a warning notice beside the existing sign-in notice. An account with genuinely no teams stays silent, which is the case that decides whether this is a useful signal or a false alarm.

`parseTeams` now counts entries it cannot read. A non-empty array that yields nothing usable is a response we could not read rather than an account with no teams, so it returns `InvalidTeamsResponse`; a partly readable list keeps what it got and still reports the shortfall, since the team you want may be the entry that dropped. Without this a `200` carrying `{"teams":[{"id":123,"slug":"acme"}]}` reproduced the original bug with no error anywhere.

That last part also reaches `fx teams`, which now fails loudly on a list it cannot read instead of reporting no teams. It already used `try` on the same call, so this is the direction it was already going, but it is a behavior change in a second command and worth your attention.

The two surfaces share a prefix constant and append their own remedy, following `remote_revocation_warning`, so the in-app notice does not tell you to run a shell command.

### On the test harness

Two changes there were load-bearing rather than incidental.

`TestAuth.openTeamPicker` was a no-op that did not take the selection the way the real picker does. The consumer reads the flag before the picker branch precisely because `openTeamPicker` takes ownership, and with the old stub a unit test covering that could pass whether or not the read was in the right place. Making the stub faithful is what lets the test actually pin it: removing the hoist now turns it red.

`TestApp` discarded `notice.tone`, so nothing could distinguish a warning notice from a neutral one. It records it now, and the test asserts the tone.

### Verification

Nine new end-to-end cases and six new unit tests. The three failure cases were each observed failing before the implementation, and both silence controls were proven load-bearing by forcing the flag true and watching them go red.

The full `tui-auth-source-selection.test.ts` run was compared against the same file on `main`: identical failing-test sets, 14 either way, none of them these. The Zig suite was compared the same way, by failing test name rather than count. `zig fmt --check src/` passes.

Full CI on my fork is green on this commit. It took a few runs to get there, and the reason is worth recording rather than hiding: `E2E (ReleaseSafe, macos-aarch64, shard 3/4)` failed twice before passing, on subagent-manager tests.

That shard is flaky on `main` independently of this change. I pushed `261a7c3f` unmodified to my fork and ran the same workflow: it passed once and then failed the same shard, on `configure rejects a concurrent winner then retries the preserved draft once`, `persistent child pointer drag replaces the selected composer range`, and `persistent child preserves its reading position across both reopen paths`. Across runs the failing set keeps changing, and the pointer-drag test failed on both `main` and this branch.

Tally on `macos-aarch64 shard 3/4`: unmodified `main` 1 pass / 1 fail, this branch 1 pass / 2 fail.

I mention it because my first read was that the shard could not be related since its tests never touch the login path, and that reasoning was wrong: every e2e test drives the same binary this change is compiled into. The A/B against unmodified `main` is what actually settles it, not the argument from test names.

### Not covered

A socket-level transport failure is not directly exercised. It reaches the same catch as the HTTP cases, so the flag is covered, but the socket path itself is not.

The wasm `fetchTeamsFromJsHost` branch has no test here.

Re-authenticating over an existing session still clears a previously selected team when the fetch fails. That predates this change and fixing it means deciding a team should survive re-login, which felt like a separate call to make.

### DIFF
--- a/src/core/app/app_auth_runtime.zig
+++ b/src/core/app/app_auth_runtime.zig
@@ -389,6 +389,9 @@ pub fn Runtime(comptime App: type) type {
                             }
                             rememberCredentialSource(app, .fx_login);
 
+                            // openTeamPicker takes the selection, so read the flag first.
+                            const team_fetch_failed = selection.team_fetch_failed;
+
                             if (selection.teams.items.len > 0) {
                                 app.auth.openTeamPicker(app.alloc, selection);
                             } else {
@@ -399,6 +402,13 @@ pub fn Runtime(comptime App: type) type {
                                 .tone = .neutral,
                                 .body = "Signed in to Vercel.",
                             });
+                            if (team_fetch_failed) {
+                                try writeAuthNotice(app, .{
+                                    .topic = "auth",
+                                    .tone = .warning,
+                                    .body = login_flow.team_fetch_warning_interactive,
+                                });
+                            }
                         },
                         .chatgpt => {
                             try finishSubscriptionSignIn(app, .codex);
@@ -1457,6 +1467,7 @@ const TestAuth = struct {
     catalog_ready: bool = true,
     gateway_ready_after_refresh_count: ?usize = null,
     team_selection: TestTeamSelection = .{},
+    taken_selection: login_flow.TeamSelection = .{},
     selected_team_adopted: bool = false,
     sign_in_url: ?[]const u8 = null,
     picker_pop_count: usize = 0,
@@ -1556,7 +1567,11 @@ const TestAuth = struct {
         return .empty;
     }
 
-    fn openTeamPicker(_: *TestAuth, _: std.mem.Allocator, _: *login_flow.TeamSelection) void {}
+    fn openTeamPicker(self: *TestAuth, alloc: std.mem.Allocator, selection: *login_flow.TeamSelection) void {
+        // Mirrors the real picker, which takes ownership of the selection.
+        self.taken_selection.deinit(alloc);
+        self.taken_selection = selection.take();
+    }
 
     fn refreshFxLoginIfNeeded(self: *TestAuth, _: std.mem.Allocator) !bool {
         self.refresh_count += 1;
@@ -1692,6 +1707,7 @@ const TestApp = struct {
     } = .{},
     model_cache_warmup_count: usize = 0,
     notice_write_count: usize = 0,
+    last_notice_tone: ?types.NoticeTone = null,
     transcript: std.ArrayList(u8) = .empty,
     test_url_opener: TestUrlOpener = .{},
     preference_write_count: usize = 0,
@@ -1703,6 +1719,7 @@ const TestApp = struct {
 
     fn deinit(self: *TestApp) void {
         self.transcript.deinit(self.alloc);
+        self.auth.taken_selection.deinit(self.alloc);
     }
 
     fn startModelCacheWarmup(self: *TestApp) void {
@@ -1715,6 +1732,7 @@ const TestApp = struct {
 
     fn writeDomainNotice(self: *TestApp, notice: types.SemanticNotice, _: bool) !void {
         self.notice_write_count += 1;
+        self.last_notice_tone = notice.tone;
         try self.transcript.appendSlice(self.alloc, notice.body);
         try self.transcript.append(self.alloc, '\n');
     }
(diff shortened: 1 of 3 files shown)


beardthelion:
This needs a `type: bug` label and I can't add one, so per AGENTS.md I'm leaving it in draft. Happy to flip it to ready once a maintainer labels it.

On CI, so it isn't a surprise: upstream checks are green. On my fork's Full CI, everything passes except `E2E (ReleaseSafe, macos-aarch64, shard 3/4)`, which fails on two subagent-manager tests (`configure duration derives its stop boundary and survives restart`, `persistent child pointer drag replaces the selected composer range`). A `macos-aarch64` shard 1/4 failure in the first pass cleared on re-run; shard 3 did not.

I don't think that shard is related to this change, and the check that convinced me is that shard 3's file list is identical on `main` and on this branch, and `tui-auth-source-selection.test.ts` is not in it. `tui-subagent-manager.test.ts` never touches the login path. Happy to be told otherwise if that shard is known-good on your runners.

beardthelion:
Correcting my earlier note on that macOS shard. I said it looked unrelated because its tests never touch the login path, which was the wrong reason: every e2e test runs the same binary this change is compiled into, so the test file's contents prove nothing either way.

I settled it properly by pushing `261a7c3f` unmodified to my fork and running the same workflow. That shard fails there too, on `configure rejects a concurrent winner then retries the preserved draft once`, `persistent child pointer drag replaces the selected composer range`, and `persistent child preserves its reading position across both reopen paths`. The failing set shifts between runs, and the pointer-drag test failed on both `main` and this branch. Tally on `macos-aarch64 shard 3/4`: unmodified `main` 1 pass / 1 fail, this branch 1 pass / 2 fail.

Full CI is green on this commit now. The PR body has the details.

PR #494, opened 2026-08-28
Report stale terminal helper diagnostics

fazxes:
## Summary

- Identify stale terminal helpers when required process-tree signal capabilities are missing.
- Return the missing capability and safe restart guidance without changing ordinary terminal failure payloads.
- Tell agents to account for live sessions before closing or restarting the helper.


fazxes:
Closing as superseded by #520. The public terminal tool and the model-facing implementation boundary changed to Shell, so this patch no longer applies. Any stale-helper recovery guidance should be handled at the current Shell boundary rather than reviving the retired Terminal contract.

PR #495, opened 2026-08-28
Publish committed file diffs over ACP

haveanicedavid:
## Problem

fx computes and retains an authoritative committed-file handoff containing the
canonical path, previous content, new content, and tool-call identity.

The ACP adapter currently discards that handoff:

- `publishCommittedFileHandoff` reports the diff as skipped.
- ACP clients receive only completion text such as
  `edited README.md (5115 bytes)`.
- Clients therefore cannot render accurate inline diffs or file-change
  summaries.

A client cannot safely recover the missing information by parsing that prose,
watching the filesystem, or taking Git snapshots. Those approaches cannot
reliably attribute changes to a specific tool call and may include pre-existing
or concurrent edits.

ACP already defines structured `diff` tool-call content with `path`, `oldText`,
and `newText`, so no protocol extension is needed.

## Change

Publish the existing committed-file handoff as an ACP `tool_call_update`
containing:

- The canonical absolute file path
- The exact previous content, or `null` for a new file
- The exact committed content
- The original tool-call ID

The subsequent completion update omits `content`. ACP content updates replace
the collection when present, so omitting it preserves the previously published
diff while still advancing the tool status to `completed`.

This uses fx's existing source of truth and adds no filesystem polling,
snapshotting, or provider-specific client behavior.

## Verification

- `zig build`
- `zig fmt --check src/`
- `bun test acp.test.ts -t 'ACP publishes committed file changes as authoritative diffs'`

The end-to-end test drives the freshly built `fx acp` binary and verifies:

- The file is committed.
- The diff update contains the canonical path and exact old/new contents.
- The diff precedes the completed update.
- Completion does not overwrite the diff content.
- stderr remains empty.

Required label: `type: bug`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ fx builds as a native binary or WebAssembly. Applications embedding fx can provi
 
 | Surface | Use |
 | --- | --- |
-| `fx acp` | Connect the native agent to editors and other Agent Client Protocol clients. |
+| `fx acp` | Connect the native agent to Agent Client Protocol clients with streamed messages, tool calls, and file diffs. |
 | `createFxAgent()` | Embed the agent core in a JavaScript host with `fx-core.wasm`. |
 | `createFxTerminal()` | Embed the interactive terminal with `fx-term.wasm`. |
 
--- a/src/acp/prompt.zig
+++ b/src/acp/prompt.zig
@@ -191,6 +191,26 @@ const AcpContext = struct {
         try self.sendUpdate(out.writer.buffered());
     }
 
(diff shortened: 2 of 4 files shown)


haveanicedavid:
Converted this back to draft per CONTRIBUTING.md.

Fork CI is waiting on maintainer **Approve and run** (GitHub Actions is `action_required` for this SHA). Please also apply `type: bug` — I cannot add labels from the fork.

I will not mark this ready until Full CI and the ship gate pass on the exact current commit.

PR #523, opened 2026-08-29
Add an opt-in WebSocket transport for Codex Responses

mjlbach:
## Summary

- add `src/gateway/websocket.zig`, an RFC 6455 client codec (handshake keys, masked frames, fragmentation, control-frame rules) over `std.Io` reader/writer streams
- add `src/gateway/openai_codex_websocket.zig`: upgrade the Codex Responses endpoint to a WebSocket (`OpenAI-Beta: responses_websockets=2026-02-06`), send the request body as one `response.create` text frame, and feed each received frame into the same Responses reducer the SSE transport uses, with identical stream limits
- cache the connection per (session, account) with a five-minute idle and 55-minute age limit, so later turns skip DNS, TCP, and TLS setup; a busy or key-mismatched entry gets a one-off connection, and a reused connection that fails before any model output is replaced by a fresh one once
- opt in with `FX_OPENAI_CODEX_TRANSPORT=websocket`; the default transport is unchanged
- a WebSocket failure before any model output falls back to SSE under the original invocation admission and latches the process to SSE so a broken proxy costs one failed handshake, not one per turn; after output has been emitted the failure is surfaced instead — a replay can never duplicate tool calls
- `request.deadline` is respected end to end: it bounds the handshake and the stream phase (via the bounded cancel watcher), surfacing `error.Timeout`
- masked server frames are rejected (RFC 6455 5.1: only clients mask)

The handshake is `std.http.Client`'s existing 101/informational path, so TLS, proxies, and the cancel watcher are inherited rather than reimplemented. Protocol and policy mirror Codex ([`codex-rs/core/src/client.rs`](https://github.com/openai/codex/blob/6478a751fde8884b2fdc76486fe23175a8e795d4/codex-rs/core/src/client.rs)) and pi ([`openai-codex-responses.ts`](https://github.com/badlogic/pi-mono/blob/853a80d26c90a14c1886f0ebb8ffaae133ca2185/packages/ai/src/api/openai-codex-responses.ts)); pi's transport selection, fallback-before-output rule, and cache limits are followed exactly. Prewarm, `previous_response_id` continuation, and zstd compression on the SSE path are follow-ups.

## Testing

- `zig fmt --check` on touched files
- `zig build test` — full suite passes; new tests cover RFC 6455 vectors and frame round trips, a loopback WebSocket server proving end-to-end streaming plus connection reuse (one TCP accept serving two requests), fallback-safe handshake rejection, and the fallback error classification
- `bun test acp.test.ts` (116 tests) and `tui-auth-source-selection.test.ts` (56 tests) pass, including two new e2e tests that drive the built binary against a Bun WebSocket loopback server: a completion streamed over one upgraded connection with zero SSE requests, and pre-output fallback to SSE when the upgrade is rejected
- live against `chatgpt.com`: single-shot and tool-using turns stream over the WebSocket; a saved session logs `codex_ws_connection reused=true` on the second step, removing the ~400 ms connect observed on fresh connections

### DIFF
--- a/src/gateway/openai_codex.zig
+++ b/src/gateway/openai_codex.zig
@@ -6,6 +6,7 @@ const stream_provider = @import("../core/agent/stream_provider.zig");
 const io_mod = @import("../core/shared/io.zig");
 const types = @import("../core/shared/types.zig");
 const gateway_client = @import("client.zig");
+const codex_websocket = @import("openai_codex_websocket.zig");
 const responses_protocol = @import("responses_protocol.zig");
 const model_tool_schema = @import("../core/tooling/model_tool_schema.zig");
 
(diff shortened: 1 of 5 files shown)


mjlbach:
This PR has been folded into https://github.com/vercel-labs/fx/pull/521

PR #525, opened 2026-08-30
Add OrcaRouter as a first-class model provider

zsanig22-dotcom:
Add OrcaRouter as a named model provider.

fx is a tiny, open, embeddable coding agent that currently routes through Vercel AI Gateway, Codex, and Grok. This PR adds `orcarouter` as a fourth first-class provider alongside them: [OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway that exposes a provider/model namespace across many models, and like OpenRouter it works as a drop-in base URL — but it also adds adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance on the same endpoint. Wiring it as a named provider means fx users can use that stack directly instead of treating it as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The provider mirrors the shared Responses-protocol streaming already used by Codex and Grok, pointed at `https://api.orcarouter.ai/v1/responses`, with a model catalog fetched from `https://api.orcarouter.ai/v1/models`. Credentials resolve from `ORCAROUTER_API_KEY` through the existing credential precedence, and `fx login orcarouter` selects the provider when the key is set (no OAuth session to create).

Verification:
- `zig build` and `zig fmt --check src/` pass; the full `zig build test` suite matches baseline (same pre-existing environment-dependent failures, no new ones; 6 new OrcaRouter tests pass).
- Live: `fx models` lists 204 OrcaRouter models from the real endpoint, and `fx ask` through the provider returned a real model reply (HTTP 200).

I'm an engineer on the OrcaRouter team. Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

### DIFF
--- a/src/builtins/providers.zig
+++ b/src/builtins/providers.zig
@@ -6,6 +6,8 @@ const openai_codex_permission_reviewer = @import("../gateway/openai_codex_permis
 const xai_grok = @import("../gateway/xai_grok.zig");
 const xai_grok_models = @import("../gateway/xai_grok_models.zig");
 const xai_grok_permission_reviewer = @import("../gateway/xai_grok_permission_reviewer.zig");
+const orcarouter = @import("../gateway/orcarouter.zig");
+const orcarouter_models = @import("../gateway/orcarouter_models.zig");
 const provider_catalog = @import("../core/auth/provider_catalog.zig");
 
 pub const native = provider_set.Set{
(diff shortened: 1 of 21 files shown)


zsanig22-dotcom:
This PR should carry the `type: feature` label (per CONTRIBUTING.md). The fork contributor does not have label permissions, so a maintainer will need to apply it.

PR #531, opened 2026-08-30
Allow closing compact command menus with Esc

### LINKED ISSUE #196: Bug: "check usage" screen cannot be closed with Esc
fx 0.0.4 · Linux x86_64 · model `zai/glm-5.2` (the default)

## What happens

The `/usage` dashboard (the screen that shows local fx token usage and spend) cannot be dismissed. The footer advertises `Esc Close`, but pressing Esc does nothing — neither a single nor a double Esc closes it. The only way out is to quit fx, which loses the in-progress session.

## What I expected

Esc should close the usage dashboard and return to the session, the way it does for other overlays in fx (and as the footer itself advertises).

## Steps to reproduce

1. Start `fx` interactively in Ghostty.
2. Type `/usage` + Enter — the Usage dashboard opens.
3. Press Esc.

Result: nothing happens, the dashboard stays open. A second Esc also does nothing. Other overlays in fx close on Esc fine, so this looks specific to the usage dashboard route.

## Environment

- **Terminal:** Ghostty (reproduced consistently)
- **tmux:** Works fine — Esc closes the dashboard as expected. The bug does not reproduce under tmux.

## Impact

The only way out is to quit fx and restart, which loses the in-progress session.

### Sensitive information

adi-IL:
When a compact command menu (such as usage, statusline, or workspace) is active, `routeActiveModalInput` intercepts raw input bytes and returns `true` for all unhandled keys.

Because raw byte `27` (`Esc`) fell through to `else => {}`, pressing `Esc` was swallowed without closing the active compact menu.

Handle byte `27` by invoking `cancelCompactCommandMenu` and requesting a footer redraw.

Fixes #196.

### DIFF
--- a/src/core/app/app_input_runtime.zig
+++ b/src/core/app/app_input_runtime.zig
@@ -1381,6 +1381,10 @@ pub fn Runtime(comptime App: type) type {
                         _ = moveCompactCommandMenu(app, menu, -1);
                         app.shell.render_requests.request(.footer);
                     },
+                    27 => {
+                        _ = cancelCompactCommandMenu(app);
+                        app.shell.render_requests.request(.footer);
+                    },
                     else => {},
                 }
                 return true;


adi-IL:
Closing to keep contribution queue tidy.

PR #535, opened 2026-08-31
Add terminal color palette option

rodrigomencos:
## Summary

  - Add an `fx` or `terminal` color palette preference to `/settings`
  - Let terminal themes supply colors through the default foreground, background, and ANSI color slots
  - Keep the existing fx palette as the default
  - Preserve external command and tool ANSI output

  This allows terminal themes to style fx without embedding theme-specific color tables.

  ## Design

  - The typed contract is `ColorPalette { fx, terminal }`
  - The preference is profile-owned and stored in `~/.fx/settings.json`
  - Project `.fx.json` cannot set the preference
  - Palette changes update fx-owned UI and newly generated output immediately and persist across launches
  - transcript and external ANSI content remain unchanged.
  - Interactive fx and terminal-formatted `fx ask` use the selected palette

  ## Verification

  - `zig fmt --check src/`
  - `git diff --check`
  - `zig build`
  - Focused palette, configuration, persistence, presentation-snapshot, nested-runtime, and transcript-cache tests
  - Exercised `./zig-out/bin/fx` under Nushell
  - Verified live UI and new-output switching in both directions without rewriting retained output
  - Verified persistence after restarting fx
  - Verified terminal mode emits default and ANSI-16 colors
  - Verified external ANSI content remains unchanged

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -126,7 +126,7 @@ Config precedence (highest wins):
 4. `<workspace>/.fx.json` (committed project defaults)
 5. Built-in defaults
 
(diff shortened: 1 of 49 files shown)


adi-IL (commented):
Tested locally on Linux x86_64 with Zig 0.16.0.

Separating `presentation_style_generation` from `full_transcript_content_revision` keeps palette switching from invalidating retained transcript bytes or clearing entry histories. The cache lookup in `CompactTranscriptSourceCache.find` checks the generation alongside columns and revision cleanly.

PR #556, opened 2026-09-01
Complete ACP tool-call reporting

fazxes:
## Summary

- Include public tool names and redacted input when ACP tool calls begin.
- Report provider-backed searches as `web_search` without exposing routing aliases.
- Publish one terminal ACP update for each provider-backed search.
- Classify `web_fetch` calls with the ACP `fetch` kind.


molefrog:
Awesome! This should be very helpful for the integration I'm building

Issue #537, opened 2026-08-31
Feature request: Fork or rewind a session from an earlier turn


andriusbug:
### Issue details

 Problem:

  Long sessions often reach a point where users want to try another approach or discard recent turns. Currently,
  resuming or recovering a session does not provide an easy way to branch from—or return to—a specific earlier turn.

  Proposed commands:

  • /fork <turn> — Create and enter a new session starting at the selected turn. Preserve the original
    session unchanged.
  • /rewind <turn> — Move the current session back to the selected turn, removing subsequent turns from
    active context.

  Suggested behavior:

  • /fork displays or returns both the source and new session IDs.
  • /rewind asks for confirmation because it changes the current session.
  • Turns can be selected through an interactive picker or stable turn identifiers.
  • Tool calls, recovery checkpoints, compaction state, and attachments associated with retained turns are preserved.
  • Neither command reverses external side effects such as file edits, commands, commits, or API calls.
  • Rewound transcript data could remain recoverable internally, rather than being deleted immediately.

  Examples:

  ─ text ───
  /fork
  /fork 12
  /rewind
  /rewind 12
  ──────────

  Possible CLI equivalents:

  ─ bash ───────────────────────────────────
  fx session fork <session-id> --at <turn>
  fx session rewind <session-id> --to <turn>
  ──────────────────────────────────────────

  Key distinction:

  • /fork is nondestructive and supports exploring alternatives.
  • /rewind changes the active history and supports abandoning an unsuccessful path.

  This would provide lightweight conversation branching while making it explicit that session history and external
  workspace state are separate.

### Sensitive information

- [x] I reviewed the information above and removed secrets and other sensitive data.

Aarya2004:
Hey! Taking a look at this, I'll post an update on this thread when I can!

Aarya2004:
Hey! I was able to get the first iteration of the PR out, its just the base implementation of the CLI variants of fork and rewind. I'm working on getting another PR (based on #560) that will add /fork and /rewind in the interactive sessions. Appreciate any feedback on the PR!

Edit: I'll get the PR out of the draft state once I have /fork and /rewind implemented!

andriusbug:
> Hey! I was able to get the first iteration of the PR out, its just the base implementation of the CLI variants of fork and rewind. I'm working on getting another PR (based on [#560](https://github.com/vercel-labs/fx/pull/560)) that will add /fork and /rewind in the interactive sessions. Appreciate any feedback on the PR!
> 
[...]

PR #560, opened 2026-09-01
Add fx session fork and fx session rewind

Aarya2004:
Part of #537. Adds the non-interactive half: branch a saved session at a turn, or drop its last turns. The `/fork` and `/rewind` slash commands build on this in #568.

```
$ fx session fork <id> --at 7        # new session holding turns 1..7, source untouched
$ fx session rewind <id> --by 2      # same session, last 2 turns dropped
```

Both take `--json`. Turn numbers are the `[turn N]` labels `fx session <id>` prints, compaction summaries included. `fork --at` is absolute because you read a turn number and branch there; `rewind --by` is relative because the thought is "undo my last two".

## Try it

```bash
gh pr checkout 560 && zig build
./zig-out/bin/fx sessions                          # pick an id
./zig-out/bin/fx session <id>                      # note the [turn N] labels
./zig-out/bin/fx session fork <id> --at 2 --json   # new id, source untouched
./zig-out/bin/fx session rewind <id> --by 1        # same id, one turn fewer
./zig-out/bin/fx --resume <new id>                 # the branch resumes
```

Without saved sessions, `cd tests/e2e && bun install && bun test session-fork.test.ts` builds real ones on a fake model and checks every path above.

## How it works

Built on the conversation-log storage now on `main`; the branch is merged up to date with it.

- `planConversationTruncation` scans `events.jsonl` and returns the byte boundary after N turns. It counts either the archive, which is what `fx session <id>` labels, or the resumed window, which is the latest compaction summary plus the turns after it and what a live shell holds (#568 uses that mode). A cut that lands inside a turn left open by a compaction checkpoint closes that turn as interrupted, the way recovery does.
- **Fork** is recovery with a chosen boundary: stage a new session, copy the result files and image snapshots the kept turns reference, install the exact byte prefix of the source log, promote it. The source is opened under its writer lock and never written; a test compares every byte before and after. The tail of `recoverSessionCopy` became `publishSessionCopy` so recovery and fork stay one code path.
- **Rewind** truncates the session's log in place at that boundary and drops a paused response along with the tail it belonged to. The dropped turns leave the transcript, so fork first if they should stay reachable. A session another process holds open answers `SessionBusy`.

## Verification

- `zig build test` on the merged branch: 8633 pass, 20 skip, 1 fail in `command_effect`'s printf planner test. That test fails identically on current `main` in this environment.
- `tests/e2e/session-fork.test.ts` 7 pass; `cli.test.ts` 119 pass.
- Unit tests cover fork and rewind across a compaction checkpoint in both counting modes, the busy lock, range errors, the no-op path leaving the log byte-identical, and paused-response clearing.
- Not covered: a fork of a session carrying image attachments or tool-result artifacts; those helpers are exercised only through recovery's tests.

Needs the `type: feature` label, which I cannot apply.

https://claude.ai/code/session_01Hjm7J6N3SL5Y62TJ3bPxwD


Aarya2004:
Hey @fazxes! I think this is ready for the runner checks! 

This is the first of two stacked PRs for #537; #568 builds on it (I don't think I can stack these unfortunately).

Full CI ran without errors on all four runners for both heads, run from my fork: 
- [#560](https://github.com/Aarya2004/fx/pull/560) → https://github.com/Aarya2004/fx/actions/runs/33593235265
- [#568](https://github.com/Aarya2004/fx/issues/568) → https://github.com/Aarya2004/fx/actions/runs/33593237447
- (and main for comparison: https://github.com/Aarya2004/fx/actions/runs/33593232787).

The PGSO workflows, are waiting on first-contributor approval, and both need type: feature. I've kept them as drafts per CONTRIBUTING until that runs, happy to mark ready whenever you prefer.

PR #586, opened 2026-09-02
Skip user settings writes on js-host surfaces

srps:
## Summary

The js-host runtime profile persists the model and permission mode through the host config store (`app_host_config_runtime`). It has no HOME directory and no user settings file. Commands that also write user preferences still attempted the HOME-based write and reported its inevitable failure on every use. On the wasm terminal, `/model` prints "Model picker: active for this process but not saved to user settings (HomeNotSet)" although the host store had saved the choice.

This change returns before the HOME-based attempt when the profile allows `js_host_config`, in the two shared helpers (`persistUserPreferences`, `persistUserPreferencesSilently`) and the inline attempt in `/sound`. Native builds are unchanged.

## Verification

- `zig build -Dwasm-surface=term` on upstream `main` `19ae8f54`.
- On a JavaScript host with a config store, `/model` switches and persists the model with no error notice; on restart the host store restores it.
- Native `fx`: `/model` still writes `~/.fx` user settings and reports failures as before.

### DIFF
--- a/src/core/app/app_commands.zig
+++ b/src/core/app/app_commands.zig
@@ -166,6 +166,9 @@ fn persistUserPreferences(
     patch: config_runtime.UserSettingsPatch,
     runtime_changed: bool,
 ) !void {
+    // js-host surfaces have no user settings file; the host config store
+    // owns preferences there, so a HOME-based write has nothing to report.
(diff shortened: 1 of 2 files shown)


srps:
Hit this while embedding it via nx.js on the Switch. I patched it on my fork, not a huge deal but it makes the UI better for these niche environments.

PR #622, opened 2026-09-03
Render Mermaid diagrams in the TUI

### LINKED ISSUE #621: Render Mermaid diagrams in the TUI
## Problem

Assistant responses frequently contain fenced Mermaid diagrams. In the TUI these are currently displayed only as source code, even when the terminal has enough room to show a useful diagram directly.

For example, this response:

```mermaid
flowchart LR
  A[Request] -->|validate| B[Response]
```

could be rendered in a sufficiently wide transcript as terminal-native Unicode art similar to:

```text
┌─────────┐  validate  ┌──────────┐
│ Request │ ─────────▶ │ Response │
└─────────┘            └──────────┘
```

A sequence diagram should receive the same treatment:

```mermaid
sequenceDiagram
  User->>Agent: Ask
  Agent-->>User: Answer
```

## Proposed behavior

- Detect completed fenced code blocks whose language is `mermaid`.
- Render supported Mermaid syntax as bounded Unicode diagrams in the normal TUI transcript.
- Use the current transcript width for layout and recompute presentation during terminal resize.
- If syntax is unsupported, input exceeds safety bounds, or the finished diagram cannot fit, preserve the ordinary Mermaid source code block instead of clipping or corrupting the diagram.
- Do not render an incomplete fence while a response is still streaming.
- Keep model output as the semantic source of truth so resizing does not mutate persisted transcript content.

The expected responsive lifecycle is:

```text
wide terminal: diagram -> narrow terminal: source fallback -> wide terminal: diagram
```

## Reference implementation

Grok Build has a high-quality terminal Mermaid implementation worth using as a design reference. Its renderer passes the available width into layout and deliberately selects a source-preserving fallback when layout reports that the diagram is too wide:

- [renderer dispatch and oversize fallback](https://github.com/xai-org/grok-build/blob/bc7f02eddd3d84085849dc19ed216f11c23b0571/crates/codegen/xai-grok-markdown/src/mermaid.rs#L58-L89)
- [bounded source fallback](https://github.com/xai-org/grok-build/blob/bc7f02eddd3d84085849dc19ed216f11c23b0571/crates/codegen/xai-grok-markdown/src/mermaid.rs#L3497-L3565)

The important UX property is graceful degradation: a diagram is visual when the TUI can carry it, and remains readable as its original source when it cannot.

## Acceptance criteria

- Common flowchart, state, class, entity-relationship, and sequence diagrams render in a wide TUI.
- Unsupported and over-budget input remains visible as ordinary source.
- A diagram that becomes too wide after a resize falls back to source, then renders again when space is restored.
- Labels cannot inject terminal control sequences.
- Parser, layout, canvas, and allocation limits bound untrusted model output.
- Deterministic tmux E2E coverage exercises the complete streaming and resize lifecycle, including process liveness and clean stderr.

acking-you:
Closes #621

## Summary

- render completed `mermaid` code fences as bounded Unicode diagrams in the TUI transcript
- support common flowchart, state, class, entity-relationship, and sequence diagram syntax
- split parsing, owned models, label normalization, layout, and canvas painting into focused modules
- preserve the ordinary source code block for unsupported, over-budget, or too-wide diagrams
- recompute presentation from semantic transcript content on resize, making diagram to source fallback and restoration reversible
- sanitize terminal control characters and bound parser, graph, sequence, label, and canvas allocations

The width-aware fallback follows the UX established by [Grok Build's terminal Mermaid renderer](https://github.com/xai-org/grok-build/blob/bc7f02eddd3d84085849dc19ed216f11c23b0571/crates/codegen/xai-grok-markdown/src/mermaid.rs#L58-L89): render the diagram when the TUI can carry it, otherwise keep the original source readable.

## Example

````markdown
```mermaid
flowchart LR
  A[Request] -->|validate| B[Response]
```
````

At a wide transcript width this renders as terminal-native diagram art. Shrinking below the diagram's required width restores the fenced Mermaid source; widening the terminal renders the diagram again.

## Verification

- `zig fmt --check src/`
- focused ReleaseSafe Mermaid suite: 19/19 tests passed
- `./scripts/check-public-surface.sh`
- deterministic tmux TUI E2E: 1/1 passed with 7 assertions

The E2E covers the complete requested lifecycle in one real model turn: an open fence stays source, the closed fence renders at 90 columns, a resize to 20 columns falls back to source, and a resize back to 90 columns restores the diagram. It also verifies one model request, a live pane, clean stderr, and a usable composer after completion.

## Classification

`type: feature`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ The current directory becomes the primary workspace. Enter a prompt, or run `/he
 
 Tool calls are expanded by default. Enable `Collapse tool calls` in `/settings`, or set `"collapse_tool_calls": true` in `~/.fx/settings.json`, to show one summary per tool-call group in the main transcript. Individual calls remain available in the full transcript with Ctrl+O.
 
+Closed `mermaid` code fences render as bounded Unicode diagrams directly in the TUI. Flowcharts, state diagrams, class diagrams, entity-relationship diagrams, and sequence diagrams are supported. Unsupported syntax and diagrams that do not fit the terminal remain visible as ordinary source code blocks.
+
 The status line hides the workspace path and Git branch by default. Enable the `Status line workspace` option in `/settings`, run `/statusline workspace`, or set it in `~/.fx/settings.json`:
 
 ```json
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -2,6 +2,16 @@
 
 fx includes material from the following projects and data sources.
 
+## xAI Grok
+
+The terminal Mermaid parser, layout, and Unicode canvas design are adapted
+from the Grok terminal Markdown renderer by xAI.
+
+Copyright 2023-2026 SpaceXAI
+
+Licensed under the Apache License, Version 2.0. The license terms are included
+in the root `LICENSE` file.
+
 ## cuelume
 
 The embedded interface sounds are derived from the sound recipes in
--- a/src/ui/render_engine/mermaid.zig
+++ b/src/ui/render_engine/mermaid.zig
@@ -1,0 +1,211 @@
+//! Terminal Mermaid renderer facade.
+//!
+//! The renderer is deliberately split at ownership boundaries:
+//!
+//!     Mermaid source
+//!          │
+//!          ▼
+//!       parser ──> owned model ──> layout ──> styled canvas ──> terminal rows
+//!          │
+//!          └── unsupported or over bounds ──> ordinary code-block fallback
+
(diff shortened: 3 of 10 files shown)


acking-you:
@fazxes @DarlingtonDeveloper Could you please review this when convenient? Mermaid rendering is isolated in the TUI renderer and leaves the existing source-block behavior unchanged for unsupported or narrow diagrams. It improves terminal readability while remaining reversible on resize, with focused tests and deterministic TUI E2E coverage. Thank you!

PR #634, opened 2026-09-04
Support slashless exit alias

bennor:
Allow dummies like me who can't stop typing `exit` to end a session that way.


src/core/app/input_submit_runtime.zig
```diff
@@ -2053,6 +2059,85 @@ pub fn directCommand(expanded: []const u8) ?[]const u8 {
     return expanded[1..];
 }
 
+/// A submission without a leading slash can only name a command when the whole
```
bennor, line 2062: @fazxes I'm not sure how good this is for Zig code. It's all vibed. Feel free to fix it.

PR #687, opened 2026-09-05
Bound temporary memory during long agent turns

### LINKED ISSUE #686: Bug: Long-running tasks exhaust tens of GiB and trigger OOM kills
### Issue details

**During actual development work, a single task running for several hours consumed tens of GiB, triggered the Linux OOM killer, and left my working environment unusable. I consider this a severe stability bug and ask that it be prioritized.** This was a real task, not a stress test.

### Environment and incident

- Ubuntu 24.04.3 LTS, x86_64; approximately 64 GB host RAM, 40 GB allocated to the Ubuntu environment, and 10 GiB swap.
- On 2026-09-05, fx alone reached approximately **35.3 GiB anonymous RSS plus 8.1 GiB swap**. Free swap reached zero and the kernel killed fx. Other services subsequently failed.
- The recovered turn contained 1,171 tool steps and 3,735 distinct tool calls, but its complete recovery snapshot was only about 13.2 MiB.

The original incident occurred on my fork. I then independently reproduced the memory-growth defect on **unmodified upstream `478960a8ab9315507e0a40d4434df71898fadf13`**.

### Deterministic upstream reproduction

Build with Zig 0.16.0 using `zig build -Doptimize=ReleaseSafe -j2`. Run the built `./zig-out/bin/fx ask --json --yolo` against a local Gateway SSE fixture, in an isolated HOME/workspace. Have the fixture request one `read_file` per step with a unique call ID, changing the roughly 1 KiB file before every read, and supply a final response after 1,000 steps. No paid model is needed.

The test capped fx address space at **2 GiB**, imposed a **600-second timeout**, disabled core dumps, and sampled only fx RSS every 100 ms. The fixture discarded request bodies.

| Model request | fx RSS |
| --- | ---: |
| 100 | 81.06 MiB |
| 200 | 268.11 MiB |
| 400 | 1,002.24 MiB |
| 480 | 1,425.74 MiB |

After **490 successful reads**, fx returned `OutOfMemory` and exited with code 1. Peak sampled RSS was **1,483.49 MiB**, while the retained execution record was only **1.37 MiB**, including about **0.50 MiB of tool output**. This bounded reproduction failed through the allocator; the real incident above triggered the OS OOM killer.

### Confirmed allocation path

- [The turn arena](https://github.com/vercel-labs/fx/blob/478960a8ab9315507e0a40d4434df71898fadf13/src/core/agent/runtime/orchestrator.zig#L4115-L4117) survives until the entire user turn ends.
- [`persistRecoveryCheckpoint`](https://github.com/vercel-labs/fx/blob/478960a8ab9315507e0a40d4434df71898fadf13/src/core/agent/runtime/orchestrator.zig#L3282-L3295) repeatedly rebuilds the accumulated execution history into that arena. Earlier temporary copies remain allocated, producing approximately quadratic cumulative growth. [Successful responses also trigger these saves](https://github.com/vercel-labs/fx/blob/478960a8ab9315507e0a40d4434df71898fadf13/src/core/agent/runtime/orchestrator.zig#L6458-L6487).
- [The persistence sink](https://github.com/vercel-labs/fx/blob/478960a8ab9315507e0a40d4434df71898fadf13/src/core/session/session_log.zig#L2315-L2335) synchronously copies the incoming checkpoint and frees its previous owned copy, but this does not release the caller's temporary reconstruction.

**Expected:** memory may grow with retained history, but obsolete checkpoint copies must be released after each save. Use an independent temporary allocator with correct copy-out ownership; do not reset the whole turn arena while messages or callbacks still borrow it. Lowering the step limit alone does not fix this defect.

The step limit is unlimited by default. A separate fixture also completed 160 unchanged inspection batches without a no-progress stop. However, the OOM reproduction above uses changing evidence, so useful long-running tasks are affected too.

Related: #483 and the still-unmerged #484. This report adds the Ubuntu incident and a reproduction on the current tested upstream revision. I have not tested #484, and without a heap snapshot cannot attribute every byte of the original incident to this one allocation path.

### Sensitive information

- [x] Reviewed and removed secrets and private session content. Only aggregate measurements and synthetic test details are included.

acking-you:
Long saved turns retain each full recovery-checkpoint reconstruction, plus provider HTTP/parser scratch, in the turn arena until the turn ends. Rebuilding a growing history repeatedly therefore accumulates obsolete copies and can exhaust memory during an otherwise ordinary task.

Give checkpoint reconstruction and each provider attempt their own temporary arena. Persistence sinks synchronously copy or serialize the checkpoint; owned provider results are deep-copied before request storage is released, including deferred usage references and failure diagnostics. Existing history, retry behavior, and step limits are preserved.

The included deterministic benchmark uses a local Gateway fixture with changing 1 KiB file reads, a 2 GiB address-space cap, and a 600-second deadline. On Ubuntu x86_64 with Zig 0.16 ReleaseSafe:

| Version | Successful tool steps | Peak fx RSS | Retained tool output |
| --- | ---: | ---: | ---: |
| Upstream `478960a8` | 482, then `OutOfMemory` | 1433.7 MiB | 0.50 MiB |
| This fix | 1000, completed | 82.2 MiB | 1.03 MiB |

Validation includes 1000 growing checkpoint saves, 1000 provider attempts covering success, cancellation and failures, allocation-failure injection for result copies, and runtime CLI/ACP/TUI checks for recovery, retry, cancellation, streaming and persistent child turns. Full CI passed on all four native platforms for `21fd1b6d`: [full run](https://github.com/acking-you/fx/actions/runs/33961697797).

Required classification: `type: bug` .

Fixes #686

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ zig build -Doptimize=ReleaseSafe
 
 Run the test suite with `zig build test`. See [CONTRIBUTING.md](CONTRIBUTING.md) for development and contribution guidelines.
 
+A [bounded long-turn memory benchmark](docs/long-turn-memory.md) exercises saved turns against a local model fixture.
+
 ## License
 
 [Apache-2.0](LICENSE)
--- a/benchmarks/long_turn_memory.py
+++ b/benchmarks/long_turn_memory.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+"""Bounded Linux RSS measurement of a saved fx turn against a local Gateway fixture.
+
+Only fx is measured. The fixture never retains requests; changing file contents
+make each read useful.
+Artifacts include RSS samples, stdout/stderr and the private session directory.
+"""
+import argparse
+import hashlib
+import http.server
+import json
+import os
+from pathlib import Path
+import resource
+import signal
+import subprocess
+import sys
+import threading
+import time
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--binary", default="./zig-out/bin/fx")
+    parser.add_argument("--output", required=True)
+    parser.add_argument("--steps", type=int, default=1000)
+    parser.add_argument("--bytes", type=int, default=1024)
+    parser.add_argument("--limit-mib", type=int, default=2048)
+    parser.add_argument("--timeout", type=int, default=600)
+    args = parser.parse_args()
+    if not sys.platform.startswith("linux"):
+        parser.error("RSS measurement requires Linux /proc")
+    binary = str(Path(args.binary).resolve())
+    binary_sha256 = hashlib.sha256(Path(binary).read_bytes()).hexdigest()
+    root = Path(args.output).resolve()
+    root.mkdir(mode=0o700, parents=True, exist_ok=False)
+    home, workspace = root / "home", root / "workspace"
+    (home / ".fx").mkdir(parents=True, mode=0o700)
+    workspace.mkdir()
+    (home / ".fx/settings.json").write_text(json.dumps({"provider": "gateway", "model": "openai/gpt-5.5", "max_agent_steps": 0}))
+    count = 0
+    request_bytes = 0
+
+    class Handler(http.server.BaseHTTPRequestHandler):
+        protocol_version = "HTTP/1.1"
+
+        def log_message(self, *_):
+            pass
+
+        def reply(self, data, content_type):
+            self.send_response(200)
+            self.send_header("Content-Type", content_type)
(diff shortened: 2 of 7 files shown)


acking-you:
@fazxes @DarlingtonDeveloper Could you please review this when you have time? This fixes a severe long-turn memory issue: repeated checkpoint reconstruction and provider scratch accumulated until turn end, so GPT-style runs spanning 1,000+ steps could OOM fx. The fix bounds temporary arenas while preserving history and retry behavior; the benchmark drops peak RSS from 1.43 GiB (482 steps, OOM) to 82 MiB (1,000 steps completed). Thank you!

PR #736, opened 2026-09-07
Preserve session ownership and report save failures

fazxes:
## Summary

- Keep saved sessions exclusively owned until close, including while suspended with Ctrl+Z.
- Roll back failed conversation writes and refuse further writes when rollback cannot be confirmed.
- Report shutdown save failures with a warning and nonzero exit, including when terminal input closes.
- Recover conversations with corrupt usage data into a new session with incomplete historical accounting, preserving the original.


fazxes:
Superseded by #743. The replacement builds on the persistence fixes now on main and keeps only the remaining session-ownership and accounting-copy changes.

PR #753, opened 2026-09-08
Keep unmatched emphasis and backtick markers literal

fazxes:
## Summary

- A `**`, `*`, `__`, `_`, or `~~` opener with no matching closer on the same line renders as the literal characters instead of styling the rest of the line.
- A backtick with no closing backtick renders literally instead of turning the rest of the line into an inline code span.
- Matched spans, spans nested across marker styles, code spans that contain markers, and backslash-escaped markers render exactly as before.

This intentionally changes the earlier behavior where an unclosed opener stayed active to the end of the line; that behavior is what made a leading `*` or `_` in prose italicize whole sentences.

Stacked on #752.


fazxes:
Superseded by #762, which replaces the lookahead approach with a delimiter stack.

PR #756, opened 2026-09-08
Saturate OAuth refresh deadline subtraction

### LINKED ISSUE #445: Bug: an out-of-range stored expiry aborts fx instead of reading as expired
### Issue details

`refreshDeadlineMs` in `src/core/auth/grok_session.zig:19` and `src/core/auth/chatgpt_session.zig:19` subtracts the one minute safety margin with a plain `-`:

```zig
pub fn refreshDeadlineMs(expires_at_ms: i64) i64 {
    return @max(expires_at_ms - expiry_skew_ms, 0);
}
```

Nothing bounds `expires_at_ms`. It comes straight out of the stored auth file: `parse` reads it through `requiredInteger` (`grok_session.zig:220` and `:249`, `chatgpt_session.zig:211` and `:239`), which accepts any JSON integer and range-checks nothing. At `expires_at_ms < i64.MIN + 60_000` the subtraction overflows and the process aborts. The `@max` reads as if it covers this, but it runs on a result the subtraction never produces.

The interesting part is the contrast with the rest of the same parser. A bad `account_id`, a string where the expiry belongs, a wrong `version`: each returns `error.InvalidGrokAuthSession`, and `loadFromDir` turns that into no session (`grok_session.zig:131-137`). This one field aborts instead.

`oauth_session.zig:106` does the same arithmetic with a saturating `-|` and is unaffected.

Reproduced on `385f74e0` with `~/.fx/grok-auth.json` written at mode `0600` under a temporary `HOME`:

```json
{"version":1,"access_token":"a","refresh_token":"r","expires_at_ms":-9223372036854775808,"account_id":"acct"}
```

`grok_session.load` returns the session, and the first thing that asks whether it is stale panics:

```
thread panic: integer overflow
  grok_session.zig:19  in refreshDeadlineMs   return @max(expires_at_ms - expiry_skew_ms, 0);
  grok_session.zig:44  in expired             return refreshDeadlineMs(self.expires_at_ms) <= now_ms;
```

Both refresh modes reach it. `loadAccess` calls `expired` at `grok_oauth.zig:508`, and `takeAccess` calls `refreshDeadlineMs` at `:522`, so `.stored` panics too without ever consulting the expiry for a decision. ChatGPT is the same at `chatgpt_oauth.zig:407` and `:421`.

Normal values are unaffected: 100000 gives 40000, 10000 and 0 and -1 all clamp to 0, and `i64.MAX` does not overflow.

For scope: this is reachable from a corrupt or hand-edited auth file, not from a hostile server. The write path is already guarded, since `expires_in` goes through `requiredPositiveInteger` and then checked `mul` and `add`, so a token response cannot produce a value in the overflow range. What it costs is that one damaged file turns every `fx` invocation that touches Grok or ChatGPT auth into an abort, where every other kind of damage to the same file degrades to a normal re-login.

The existing tests cover 100000 and 10000 only (`grok_session.zig:289-290`, `chatgpt_session.zig:267-268`), which is why the boundary was never exercised.

### Sensitive information

adi-IL:
## Why
Corrupt or hand-edited Grok/ChatGPT auth files can store an out-of-range `expires_at_ms`. `refreshDeadlineMs` used an unchecked `-` for the one-minute skew, so values near `i64.min` overflowed and aborted fx. Other parse failures on the same file degrade to a re-login. This one field crashed instead.

## Scope
- `src/core/auth/grok_session.zig` and `src/core/auth/chatgpt_session.zig`: use saturating subtraction (`-|`), matching `oauth_session.zig`.
- Extend unit tests for `0`, `-1`, `i64.min`, and `i64.max`.

## Blast radius
Only the refresh-deadline helper for Grok and ChatGPT sessions. Healthy token write paths already use checked math and are unchanged.

## Verification
- Unit tests added for the overflow boundary.
- Could not run `zig build test` or `./zig-out/bin/fx` in this environment (no local Zig toolchain / Full CI). Please treat Full CI on this draft as the authoritative proof.

Required classification: `type: bug`.

Fixes #445


adi-IL:
Marked ready for review. GitHub Actions on this fork PR is sitting in `action_required` (0 runs), so Full CI has not started. If a maintainer can approve the workflow runs, that should unlock the real test suite.

PR #780, opened 2026-09-09
Add configurable model connections

fazxes:
## Summary

- Configure named Chat Completions connections in profile settings, with anonymous local access or bearer authentication from a named environment variable.
- Select connections with `fx provider` or `FX_PROVIDER`, and override model IDs with `FX_MODEL` without requiring catalog access.
- Use the selected connection for interactive chat, `fx ask`, native ACP, subagents and compaction.
- Apply the connection's configured model limits to subagent requests and compaction.
- Run automatic permission review on the same connection, with an optional reviewer model.
- Preserve saved connection bindings and reject changed destinations or credential slots on resume while keeping history readable.
- Keep project configuration from defining or selecting model endpoints.
- Fail invalid connection definitions before model requests instead of falling back to Gateway.
- Keep local inspection available for unsafe profile directories, with model requests disabled until the profile is repaired and fx restarts.
- Show the configured endpoint in status and expose profile model metadata without discovery requests.
- Stream reasoning from compatible Chat Completions servers alongside text and function calls.
- Preserve plain, signed and encrypted reasoning metadata through tool turns and saved-session continuation.
- Keep reasoning metadata bound to its original connection, model and associated tool calls.
- Continue after reasoning-only assistant turns without rejecting the next request.
- Report observed token usage without treating unknown cost as zero.
- Accept cumulative and partial usage updates without double counting or accepting contradictory counters.
- Omit tool controls when no tools are advertised.
- Disable Gateway-only search, credits and fallback vision on configured connections.
- Accept matching empty terminal usage trailers while rejecting incomplete or contradictory tool streams.
- Cancel stalled requests without replaying their POST.
- Refuse credential-forwarding redirects.


fazxes:
Local checks for the reasoning compatibility update passed: 68 focused Zig tests, 39 configured-provider E2Es, and 6 configured-provider ACP tests.

Live checks covered Ollama Qwen 2.5 and Qwen 3, plus OpenRouter Qwen's reasoning, file-tool and saved-session continuation flow. OpenRouter GPT-4.1 mini and GPT-5 mini replies also passed; GPT-5 mini retained encrypted reasoning metadata.

Live vLLM coverage and the remaining OpenRouter tool/resume/terminal routes are still unverified. This is not a claim that every model or server configuration works. Full CI and the final readiness check remain pending for this commit.

PR #781, opened 2026-09-09
Normalize root search paths in file evidence

### LINKED ISSUE #779: Bug: Empty grep_files path can make a compacted session unresumable
### Description

A successful `grep_files` call with an explicitly empty optional `path` can add invalid file evidence to the active turn. The search itself succeeds because the tool decoder normalizes an empty path to `"."`, but the durable execution-memory projection copies the raw empty string into `FileEvidence.path`.

The invalid evidence remains latent while the turn is active. If context compaction checkpoints that turn, later turn finalization or recovery materialization rejects the empty path with `InvalidConversationEvent`. Shutdown can retry the same persistence path and report a second error:

```text
fx: InvalidConversationEvent
fx: session save failed: InvalidConversationEvent
```

The conversation event log and compaction checkpoint can still be valid, so `fx doctor` and `fx session --id <id> --json` may both succeed even though continuing or finalizing the recovered turn fails.

Observed with fx 0.0.8 on Arch Linux.

### Steps to reproduce

1. Start a saved session and keep one turn active.
2. Execute a successful root-scoped search whose raw arguments contain an empty path:

   ```json
   {"pattern":"needle","path":""}
   ```

3. Trigger context compaction while that turn is active.
4. Let the turn finish, interrupt it, or resume it and submit a fresh prompt.
5. Observe `InvalidConversationEvent` while fx persists the completed or interrupted turn. On shutdown, observe `session save failed: InvalidConversationEvent` as the same state is retried.

### Expected behavior

Root-scoped searches should store a valid effective path, matching the path used by the tool. Compaction and recovery should not make the session impossible to finalize or resume.

### Actual behavior

`grep_files` execution treats `"path":""` as `"."`, while `appendFileEvidenceForTool` stores the raw empty string. `validateConversationPath` later rejects that value because conversation paths must be non-empty.

### Root cause

The producer and validator disagree:

- `src/tools/filesystem/grep_files.zig` normalizes an omitted or empty search path to `"."` for execution.
- `src/core/agent/execution_memory.zig` derives file evidence from the raw tool-call arguments and accepts an empty string.
- `src/core/session/session_event.zig` rejects empty `FileEvidence.path` values when finalizing the conversation event.

In the affected session, the final saved event was a valid context checkpoint. The invalid values existed only in the recovery checkpoint's execution file evidence, which explains why ordinary session inspection did not detect the failure.

### Suggested fix

Normalize missing and empty search roots to `"."` when creating file evidence, and add regression coverage for both explicit empty paths and omitted paths.

Existing affected sessions may still require recovery-file repair or a dedicated normalization path; preventing new invalid evidence does not retroactively change saved recovery checkpoints.

### Trace

No trace is attached because it may contain private conversation and workspace data. The diagnosis above was reproduced from structural session metadata without including prompt, output, or file contents.

### Sensitive information

mattsbl:
**Issue:** Fixes #779

## Summary

- Treat empty search-path fields as absent when projecting tool calls into durable file evidence.
- Record `"."` when a `grep_files` or `glob_files` call omits its search root or supplies an empty root.
- Add regression coverage for explicit empty and omitted search paths.

## Problem

The search tools execute an omitted or empty path against the current workspace by normalizing it to `"."`. File-evidence projection previously read the raw tool-call arguments instead and persisted `FileEvidence.path = ""`.

That empty path is rejected later by conversation-event validation. The failure can remain hidden during an active turn, become part of a recovery checkpoint after compaction, and then surface as `InvalidConversationEvent` when the completed or interrupted turn is persisted. Shutdown can retry the same invalid state and emit `session save failed: InvalidConversationEvent`.

## Changes

`appendFileEvidenceForTool` now:

1. Ignores empty `path`, `directory`, and `query` string candidates.
2. Uses the first non-empty search root when one is present.
3. Falls back to `"."` for root-scoped `grep_files` and `glob_files` calls.
4. Continues to omit file evidence for non-search tools whose required path is missing or empty.

The focused unit test covers:

- `grep_files` with `"path":""`.
- `glob_files` with no `path` field.

Both cases must produce non-empty file evidence with path `"."`.

## Scope

This change prevents new invalid file-evidence records. It does not migrate recovery checkpoints already containing an empty path. The reported local session was repaired separately by changing only the affected empty search paths to `"."` after first creating a full backup.

## Verification

Passed locally:

```text
zig fmt --check src/core/agent/execution_memory.zig
```

```text
zig test -target x86_64-linux-musl -lc -ODebug \
  --dep build_options -Mroot=src/main.zig \
  -Mbuild_options=<generated-options.zig> \
  --test-filter 'search file evidence normalizes missing and empty roots'
```

Result after merging the latest `main`: 14 filtered test roots passed, exit 0.

A patched static debug binary was built successfully for `x86_64-linux-musl`. With the separately repaired recovery file, that binary:

- inspected the affected original session with exit 0 and empty stderr;
- resumed the exact original session ID;
- displayed the preserved paused recovery at attempt 1/10;
- exited cleanly with exit 0 without making a model request.

`git diff --check` also passed with exit 0.

## Local verification limitations

The post-merge local musl Debug command below completed with exit 1 after 8,770 tests passed and 20 were skipped:

```text
zig build test -Dtarget=x86_64-linux-musl
```

The ANSI-span test launches a nested native `zig test` and failed because this Arch Linux host's GCC 16 `.sframe` startup object uses `R_X86_64_PC64` relocations that Zig 0.16's linker does not handle. Two process-disappearance timing tests also failed during the broad local run; both live in `command_runner.zig`, outside the PR's one-file diff:

- `foreground session owner loss kills the target and descendant before delayed effects`
- `timeout terminates foreground process group descendants`

The previous canonical `main` commit, `cce01969`, passed the repository's native Ubuntu ReleaseSafe CI run 34414854124. The branch now also includes current canonical `main` at `4f9b152f`; no Actions run for that commit was indexed at update time. Full CI has not yet run for the PR's exact commit, `210e529b`, and remains required on the repository's native Linux and macOS matrix.

## Privacy

No private transcript, prompt text, model output, trace, terminal recording, credential, session identifier, or unrelated workspace content is included in this PR.


fazxes:
Thank you for the patch, Matt. We verified it end to end against the crash it fixes: a turn whose search call carries an empty path made conversation validation reject the finished turn, and fx exited while saving it. With this change the session persists correctly.

This PR could not merge directly because main requires signed commits, so the identical patch landed as #936 with your credit in the commit message. Closing this one as superseded; the fix closes #779.

PR #805, opened 2026-09-11
Shrink compact ReleaseSmall binaries below 5 MiB

alvgeppetto:
## Summary

Shrinks the compact ReleaseSmall surface back under 5 MiB on aarch64-macos (4.966 MiB) with no feature changes, recovering the ~140 KiB that recent main added since the original 4.987 MiB reference, via three size levers.

- **std.json funnel:** concrete wire-type parse sites go through the shared `std.json.Value` parser plus `parseFromValue` via the new `json_owned.parseOwned` helper. All 115 typed `innerParse` parser clones (~94 KiB) are gone from the binary. Strictness is preserved: `ConversationInterruption` gains a `jsonParseFromValue` mirror of its custom hook, and the recovery checkpoint frame keeps its strict two-field shape checks. The image-measurement site in `prompt_context.zig` intentionally keeps typed parsing to borrow multi-MB payload strings from the request body.
- **`compact-release.sh --mergefunc`:** whole-program bitcode, LLVM `mergefunc`, `-Oz` recompile, ld64 link; ld64's cstring merge dedups `__TEXT,__cstring` constants the Zig linker keeps verbatim. macOS-only: Linux targets measure ~40 KiB worse and fall back to the plain build with a note. Without the flag the script behaves exactly as before.
- **simple_panic outside Debug:** stripped release binaries carry no symbols for the full panic handler's self-info reader; Debug builds keep full stack traces.

Measured and rejected, documented in CONTRIBUTING: full `-Oz` re-pipeline (+46 KiB), iroutliner (+100 KiB), RecordWire hand-walk (exact wash), `-falign-functions=4` (+16 B), mergefunc on Linux (+40 KiB).

## Numbers (aarch64-macos, stripped ReleaseSmall)

| | bytes | MiB |
| --- | --- | --- |
| Baseline after syncing with main | 5,311,656 | 5.066 |
| + std.json funnel | 5,262,072 | 5.018 |
| + --mergefunc pipeline | 5,207,696 | 4.966 |

xz download artifacts are at or under 2.45 MiB on all four targets.

## Validation

- `zig build test` green; the only 5 failures reproduce identically on clean `origin/main` (environmental, pre-existing)
- `zig fmt --check src/` and `shellcheck` clean
- Compact binaries exercised on version, help, error-path, sessions, and replay CLI paths

Label: `type: improvement` — I cannot apply labels on vercel-labs/fx, so this stays draft until a maintainer labels it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -474,6 +474,7 @@ jobs:
           scripts=(
             benchmarks/startup.sh
             scripts/check-public-surface.sh
+            scripts/compact-release.sh
           )
           bash -n "${scripts[@]}"
           shellcheck "${scripts[@]}"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -507,6 +507,50 @@ node benchmarks/libfx/bench-competitive.mjs --server /tmp/libfx-bench-server --p
 Build the SDK artifacts and install the pinned Pi package first, as shown in
 `.github/workflows/bench.yml`. Raw per-prompt samples remain in the output directory.
 
+## Compact Builds
+
+`scripts/compact-release.sh` builds stripped ReleaseSmall binaries per
+platform. It exists for size-sensitive distributions; the default product
+build remains ReleaseSafe and the macOS arm64 PGSO qualification remains the
+authoritative production pipeline.
+
+```bash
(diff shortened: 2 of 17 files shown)


alvgeppetto:
Superseded: this work is being re-posted from the videlalvaro fork with corrected commit authorship.

PR #889, opened 2026-09-15
fix: tolerate finish_reason-less terminal usage trailer in Chat Completions reducer

email2kiru:
Some OpenAI-compatible proxies send the trailing usage-only SSE frame as a populated `choices[0]` with an empty `delta` and no `finish_reason` key, instead of an empty `choices: []` array. The reducer's strict `InconsistentFinishReason` check for repeated terminal choices rejected this, breaking plain responses and tool-call turns for configured Chat Completions connections against those proxies.

This tolerates a missing `finish_reason` on an already-finished stream's trailing frame, while still requiring an exact match when it's present, and still rejecting any unexpected delta content on that frame.

Includes a regression test covering both cases.

Example trailer that currently fails (redacted):
```
data: {"id":"...","choices":[{"index":0,"delta":{}}],"usage":{"completion_tokens":6,"prompt_tokens":8,"total_tokens":14,...}}

data: [DONE]
```

The prior frame already carried `finish_reason:"stop"`; this trailer omits it entirely.


email2kiru:
Marked as draft per CONTRIBUTING.md's verification workflow — will mark ready once Full CI is green. Note: CI, Benchmarks, and Binary Size workflows currently show `action_required`, likely pending maintainer approval for a first-time contributor. Locally: `zig build`, `zig build test`, and `zig fmt --check` all pass on this commit. Suggested label: `type: bug`.

email2kiru:
@fazxes marking ready for review. CI/Benchmarks/Binary Size are still showing `action_required` — this looks like it needs a maintainer approval to run for a first-time contributor. Small, isolated fix (34 lines, one file) with a regression test; `zig build`, `zig build test`, and `zig fmt --check` all pass locally. Happy to make any changes you'd like.

Issue #109, opened 2026-08-18
provider_error: System message must be at the beginning (retries, skills discovery issues banner)


alexkreidler:
## Issue details

On the first turn of an interactive session (user message: `sup`), the provider request failed with:

```
Provider unavailable · provider_error: System message must be at the beginning. · retrying request in 2s · attempt 3/10
```

The retry loop continued (attempt N/10). Around the same time, the session banner showed:

```
Skills: 4 discovery issues; some skills may be missing (ctrl o to view)
```

and the system prompt included a `skill_discovery_warning` with `skipped_candidate_count="4"`.

**Expected behavior:** fx assembles a valid message list for the provider (system message first, or mapped to the provider's system param) so the first turn succeeds without a retry loop.

## Environment

- fx v0.0.3
- macOS (Darwin 24.6.0, Apple Silicon)
- Shell: fish (`/opt/homebrew/bin/fish`)
- Model: qwen3.8-27b (`auto · qwen3.8-27b`)
- Permission mode: auto

## Steps to reproduce

1. Have 4 skill-discovery issues on startup (banner shows "Skills: 4 discovery issues").
2. Start fx in a workspace where the skill catalog is also being truncated (`skill_catalog_bytes` context limit active).
3. Send any first message (e.g. `sup`).
4. Observe `provider_error: System message must be at the beginning` and the retry loop.

## Hypothesis

The skill-discovery warning / truncation notice is being injected ahead of the system message, or the system message isn't hoisted into the provider's dedicated system field for this backend, so providers that require a leading system message reject the payload.

---

I reviewed this report and removed secrets and other sensitive data. Happy to attach a `/trace` from a repro session.

DarlingtonDeveloper:
Thanks—this is plausible but not yet reproduced. Initial request construction currently places the system and skill context before the user message, although later execution paths can introduce system notices after conversational messages. Could you attach a sanitized /trace, including the exact model/provider and failing request role sequence, and confirm whether it reproduces without the skill-discovery warnings?

Rattlyy:
Bump getting the same issue

PR #892, opened 2026-09-15
Restore Qwen Gateway continuations

pcomans:
## Summary

- Combine multiple leading instruction blocks into one system message for canonical `alibaba/qwen` Gateway models. Preserve instruction order, original arrays, and saved history. Zero or one instruction and other model families remain unchanged.
- Add focused regressions for tool-free saved-session continuation, tool continuations, model switching, source immutability, and invalid late system messages.
- Document the compatibility policy and its measured scope.

Related to #109.

## Scope

The PR diff now contains only the Qwen compatibility fix, its regression tests, and concise documentation. Investigation-only tracing, raw-response dumping, diagnostic tests, and the research diary have been removed from the diff. The existing HTTP client and existing fx tracing are unchanged from the branch base. The requested research handoff is preserved locally, not shipped.

## Evidence

Live tests through the built fx binary reproduced the Jinja system-position HTTP 400 on ordinary second turns and post-tool continuations with multiple leading system blocks. Combining those blocks succeeded in repeated A/B tool tests and text continuations through Fireworks. The final implementation also passed live tool and text-continuation checks.

The exact downstream Gateway/template transformation remains unobserved. Family-wide selection is a compatibility policy supported by the qwen3.8-max/Fireworks reproduction, not a claim that every Qwen route was independently verified. No new SDK or provider abstraction is introduced.

## Verification after removing diagnostics

All checks passed with exit 0:

- `zig build`
- `zig fmt --check src/`
- `git diff --check` and `git diff --cached --check`
- `./scripts/check-public-surface.sh`
- Focused Zig selection with `--test-filter Qwen`: four behavior tests plus 13 discovery blocks.
- From `tests/e2e`: `FX_E2E_DISABLE_DOTENV=1 bun test ./gateway-stream-lifecycle.test.ts --test-name-pattern "Qwen Gateway continuations"`: two passed, 93 assertions. Tests inspect captured request payloads directly and do not use tracing.
- Separate direct two-turn run of the rebuilt binary against a local fake Gateway: both exits 0, no signals, empty stderr, no tool calls, and the expected `system,user,assistant,user` request sequence.

The existing E2E file retains its PGSO training classification. Adversarial review findings were addressed before the diagnostic cleanup.

## Draft gates

Required classification: **type: bug**. The author has read-only upstream permissions; a maintainer must apply the label and approve fork workflows. Keep this PR in draft until the exact-commit Full CI and final ship gate requirements are satisfied. No merge or release action is requested.


pcomans:
## Investigation results — closing in favour of a broader fix

I ran four parallel investigations against this PR: live determinism runs, a live matrix across 17 Gateway model ids with both the unfixed (`main`) and fixed (this branch) binaries, a survey of 75 Hugging Face chat templates, and a cache measurement. Summary below; the conclusion is that the `alibaba/qwen` allowlist is the wrong shape, so I'm closing this and will follow up with a differently scoped change.

### What causes the 400

Qwen 3.5, 3.6 and 3.8 templates contain, in the main message loop ([Qwen3.5-27B L83-86](https://huggingface.co/Qwen/Qwen3.5-27B/blob/main/chat_template.jinja#L83-L86), [Qwen3.6-27B L83-86](https://huggingface.co/Qwen/Qwen3.6-27B/blob/main/chat_template.jinja#L83-L86), [Qwen3.8-27B L104-107](https://huggingface.co/Qwen/Qwen3.8-27B/blob/main/chat_template.jinja#L104-L107), [Qwen3.8-2.4T-A95B L81-84](https://huggingface.co/Qwen/Qwen3.8-2.4T-A95B/blob/main/chat_template.jinja#L81-L84)):

```jinja
{%- if message.role == "system" %}
    {%- if not loop.first %}
        {{- raise_exception('System message must be at the beginning.') }}
```

Any system message not at index 0 raises, so fx's second leading system block fails the same way a late one would. Pre-3.5 Qwen templates ([Qwen3-32B](https://huggingface.co/Qwen/Qwen3-32B/blob/main/chat_template.jinja), Coder, Next, VL-Thinking) have no check. Fireworks applies the model's template and wraps the raise as `jinja template rendering failed. System message must be at the beginning.`

### Why the first turn passed and later turns failed

Not the message count: fx's first turn already carries 7 leading systems. Isolation runs against `alibaba/qwen3.8-max` (all routed to Fireworks, 8/8 first turns passed, 5/5 resumes failed, fully deterministic):

| Crafted request | Result |
|---|---|
| `[system×7, user, assistant(text only), user]` | 200 |
| `[system×7, user, assistant(reasoning + text), user]` | 400, same jinja error |
| `[system×1, user, assistant(reasoning + text), user]` (this branch) | 200 |

The trigger is a `reasoning` part in assistant history. The AI SDK converter emits it as `reasoning_content`, which pushes Fireworks onto a strict render of the template while plain requests take a lenient path. Issue #109's first-turn failure on `qwen3.8-27b` is consistent with that id's routes applying the template strictly on every turn.

### It is not Qwen-only, and it is route-specific

Rendering fx's `[system, system, user]` layout through 75 templates:

| Behaviour on a second leading system | Families |
|---|---|
| Hard raise | Qwen 3.5 / 3.6 / 3.8 open weights; Mistral Nemo, Codestral |
| Raise only for a mid-conversation system | Ministral-3, Mistral Large 3, Mistral Medium 3.5 |
| **Silently dropped** | MiniMax M2–M3, openai/gpt-oss-*, Nemotron-3 Nano, Qwen3-VL-Instruct |
| Rendered inline (fine) | Qwen3 / Coder / Next, DeepSeek, GLM, Kimi, Llama, Gemma 4, Nemotron Super/Ultra, and the rest |

Silent drops mean project instructions or runtime context vanish with a 200. Alibaba's own DashScope docs state the same contract: "Only `messages[0]` supports the system role."

Meanwhile the same family through other routes is fine: `alibaba/qwen3-coder` (Vertex) and `alibaba/qwen-3-32b` (Bedrock) accepted 7 leading systems on every turn. Six non-Qwen families also accepted them live. The constraint belongs to the (route, template) pair, and nothing in the model id or the `/v1/models` catalog tells you the route.

### Merging everywhere is cache-neutral

4-turn conversations through the Gateway with `caching: "auto"`:

| turn | layout | model | inputTokens | cacheRead | cacheWrite |
|---|---|---|---|---|---|
| 2 | 7 blocks | sonnet-4.6 | 8740 | 8665 | 72 |
| 2 | 1 block | sonnet-4.6 | 8735 | 8671 | 61 |
| 4 | 7 blocks | sonnet-4.6 | 8800 | 8776 | 21 |
| 4 | 1 block | sonnet-4.6 | 8795 | 8771 | 21 |
| 2–4 | 7 blocks | gpt-5 | 5868 / 5963 / 5988 | 5760 / 5760 / 5888 | — |
| 2–4 | 1 block | gpt-5 | 5851 / 5962 / 5987 | 5760 / 5760 / 5888 | — |

fx sets no cache breakpoints itself; the Gateway's auto breakpoints land after the whole system region regardless of block count. The `ephemeral_overlay` lane is byte-identical across turns (it only changes when the UTC date, workspace or permission mode changes).

### Conclusions

1. The merge in this PR is correct and lossless, and it fixes qwen3.8-max end to end including tool continuations.
2. An allowlist keyed on model id under-matches (Mistral, the silent-drop families), over-matches (15 pre-3.5 Qwen ids), and cannot track routes. Silent drops also rule out any error-driven fallback.
3. The Gateway already has an internal system-message consolidation step for a few providers that hoists late systems but does not merge leading ones and does not cover Fireworks. A Gateway-side fix would cover every client; I'll raise it there.
4. For fx, the data supports merging leading system messages for every Gateway model (or, as a conservative step, everywhere except Anthropic/OpenAI/Google, which is what another Vercel product already does). Follow-up PR to come.

### Not verified

`qwen3.8-27b`, `qwen3.6-27b`, `qwen3.5-plus`, and the Mistral/Gemma/Xiaomi ids returned 403 from the account used (provider restrictions), so their live behaviour is inferred from templates only. Fireworks' internal render paths are closed; the reasoning-part mechanism is inferred from the isolation runs. Cache neutrality was measured on Sonnet 4.6 and gpt-5 only.

Sources: [Qwen3.6-27B chat template](https://huggingface.co/Qwen/Qwen3.6-27B/raw/main/chat_template.jinja) · [DashScope OpenAI compatibility](https://www.alibabacloud.com/help/en/model-studio/compatibility-of-openai-with-dashscope) · [Fireworks querying text models](https://docs.fireworks.ai/guides/querying-text-models) · vllm-project/vllm#41114 · NousResearch/hermes-agent#20866 · BerriAI/litellm#39282, #39852

pcomans:
Closing: the fix is correct but scoped to the wrong predicate. See the investigation comment above; a follow-up with broader scope will replace this.

pcomans:
Decision: fx will not ship a client-side merge. The constraint belongs to the (route, template) pair, which only the Gateway knows, and the Gateway already has a `consolidateSystemMessages` step it can extend. Tracking the fix in vercel/ai-gateway#3967.

For reference, a denylist variant of this change (merge for every model except `anthropic/`, `openai/` other than gpt-oss, and `google/`) is pushed to `pcomans/fx` branch `fix/gateway-single-system-message`. It passed the focused e2e cases and a live 3-turn check against qwen3.8-max, kimi-k3 and gpt-5.5, but it is not being opened as a PR.

The investigation harness is in #901.

PR #983, opened 2026-09-20
Reduce input runtime initialization overhead

fazxes:
## Summary

- Initialize input and picker state directly into application storage.
- Preserve every declared default while leaving scratch query buffers uninitialized as designed.
- Keep one-time default stores out of the profiled interactive entrypoint.
- Remove the large read-only initializer and startup copy generated for the input runtime.

## ReleaseSafe size

| Target | Base | Candidate | Delta |
|---|---:|---:|---:|
| macOS arm64 | 10,802,544 | 10,786,128 | -16,416 bytes |
| macOS x86_64 | 12,812,041 | 12,803,849 | -8,192 bytes |
| Linux arm64 | 10,388,584 | 10,381,608 | -6,976 bytes |
| Linux x86_64 | 12,249,800 | 12,243,032 | -6,768 bytes |

On macOS arm64, read-only constants decrease by 7,840 bytes. The outlined initialization leaves `__TEXT` 6,976 bytes smaller and crosses one file-allocation boundary.

## Verification

- `zig build test`: 9,300 passed, 2 skipped
- Focused TUI startup, file picker, composer, navigation, and slash-menu E2E: 146 passed, 1 environment-dependent skip
- Interactive terminal performance gate: 50 measured samples per action passed
- ReleaseSafe builds for macOS and Linux on arm64 and x86_64

Full CI and PGSO qualification are pending for the latest SHA.


fazxes:
Closing because the exact qualified parent comparison did not reduce the macOS arm64 PGSO file, and the calibrated paired gate found a resolved normal-flow RSS regression: +99,287 bytes mean, with a 95% interval from +55,214 to +144,015 bytes. This does not meet the no-regression requirement.

PR #985, opened 2026-09-20
Compact syntax highlighting profiles

fazxes:
## Summary

- Reduce syntax profile metadata stored in the native binary.
- Preserve every language, alias, detection rule, token, and rendered style.


fazxes:
Closing this approach. It reduced the macOS arm64 PGSO binary from 6,707,536 to 6,691,120 bytes, a 16,416-byte decrease, but the paired comparison resolved RSS regressions of 113,377 bytes for normal requests and 156,140 bytes for recovered requests. Native footprint remained neutral, but the RSS increase fails the memory-preservation requirement.

